PR #1, opened 2019-11-23
fixes show_results grid bug

lgvaz:
Because `get_grid` always return a number of subplots rounded **up** there could be a mismatch between the number of items to display and the number of subplots, and the following error would be thrown for example:
```python
<ipython-input-19-a0d8e279dfe2> in show_results(x, y, samples, outs, ctxs, max_n, rows, cols, figsize, **kwargs)
      4     if ctxs is None: ctxs = get_grid(min(len(samples), max_n), rows=rows, cols=cols, add_vert=1, figsize=figsize, double=True)
      5     for i in range(2):
----> 6         ctxs[::2] = [b.show(ctx=c, **kwargs) for b,c,_ in zip(samples.itemgot(i),ctxs[::2],range(max_n))]
      7     for x in [samples,outs]:
      8         ctxs[1::2] = [b.show(ctx=c, **kwargs) for b,c,_ in zip(x.itemgot(0),ctxs[1::2],range(max_n))]

ValueError: cannot copy sequence with size 10 to array axis with dimension 12
```

In the proposed change I just make sure we don't have more subplots than items. Since the number of subplots is always rounded up we never have the problem of having less subplots than items.


sgugger:
Thanks for your PR! I think the change should be made in `get_grid` (defined in notebook 09) and not here. Just add the [:n] after `ctxs` in the return statement there. Since it's in another notebook, I'm closing this one and will let you open another one changing notebook 09. 

If you can't do it, let me know and I'll make the change when I have time.

PR #2, opened 2019-11-24
Fixes get_grid size mismatch bug

lgvaz:
Apply the changes discussed in #1 

I tried doing the change in a more compact way. The other approach would be to do `axs = axs[:n]` after line 79.

Let me know if you think that is a better idea.


lgvaz:
I also see that in the files I'm modifying there's the change:
```
   "execution_count": null,
   "execution_count": 3,
```
Does that mean my git hooks did not get setup correctly?

sgugger:
Yes, you probably forgot to run `tools/run-after-git-clone` after cloning the repo and the automatic stripout isn't done.

sgugger:
Thanks for the PR, it's much cleaner this way. Please run the script for next time, I'll stripout manually this one.

PR #4, opened 2019-11-27
Adds InstanceNorm

lgvaz:
While playing around with fast style transfer I had to use `InstanceNorm`, the support was not there so I added it =)

For some reason there is a bunch of notebook metadata change on `nbs/03_torch_core.ipynb` but I did ran `nbdev_install_git_hooks` I swear lol.


sgugger:
Looks good to me, thanks!

sgugger:
@lgvaz Didn't catch it when reviewing but as a follow-up, your norm2 test for instance norm was making the one for batchnorm useless. Fixed now :)

lgvaz:
Wow, that was a very sneaky bug hahaha, thanks for letting me know! =)

PR #5, opened 2019-11-27
Adds norm_type parameter to XResNet

lgvaz:
Adds the possibility to select a NormType when creating your XResNet (possibly `InstanceNorm` when combined with #4 )

Also fix a couple of bugs:
* Pass act_cls to stem
* `sa if i==blocks-1` instead of `sa if i==blocks`


sgugger:
Thanks!

PR #9, opened 2019-11-30
fixes norm init

lgvaz:
`_get_norm` was raising an error when `affine=False` because `bn.weight` is `None`. We should only initialize weights if `affine=True`.


sgugger:
Thanks!

PR #11, opened 2019-11-30
fixes in_channels

lgvaz:
Calling `in_channels` would fail if the first layer was a `Batch/Instance Norm` layer because this layers have an attribute called `weight`. If `affine=True` then the error was raised:
```python
~/forks/fastai2/fastai2/layers.py in in_channels(m)
    563     "Return the shape of the first weight layer in `m`."
    564     for l in flatten_model(m):
--> 565         if hasattr(l, 'weight'): return l.weight.shape[1]
    566     raise Exception('No weight layer')

IndexError: tuple index out of range
```

If `affine=False` the layer would still have `weight` but it would be None. One way of solving the issue would be to assert that the layer we are testing is a constitutional layer, but I found that would be too restrictive, maybe the user implements his own ConvLayer. So I check that the weights should have 4 dimensions, I think all convolutions should satisfy that.

---
Off topic:
One of the test depends on #9 being approved, so I commented it out. When I want to make multiple changes, should I create a branch and a separate PR for each of then or should I do everything together in a single branch?

### DIFF
--- a/fastai2/layers.py
+++ b/fastai2/layers.py
@@ -562,5 +562,6 @@ class NoneReduce():
 def in_channels(m):
     "Return the shape of the first weight layer in `m`."
     for l in flatten_model(m):
-        if hasattr(l, 'weight'): return l.weight.shape[1]
+        if getattr(l, 'weight', None) is not None and l.weight.ndim==4:
+            return l.weight.shape[1]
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks!
You should definitely make separate PRs for different fixes.

PR #17, opened 2019-12-04
nbdev export eu_langs for SentencePieceTokenizer

ismailfatih:
nbdev export flag was missing for eu_langs, causing NameError: name 'eu_langs' is not defined error for SentencePieceTokenizer


sgugger:
Thanks!

PR #21, opened 2019-12-11
Add the git clone step

krishnakalyan3:


sgugger:
HI there! You need to update the notebook index.ipynb, not the README. Since the README is autogenerated from this notebook, your change will be erased the next time we build the docs.

Could you make another PR changing the index? Thanks in advance!

krishnakalyan3:
Thanks @sgugger. I will make the change.

PR #23, opened 2019-12-12
Removing se_resnet default p value

bearpelican:
Dropout module doesn't like None p values
https://pytorch.org/docs/stable/_modules/torch/nn/modules/dropout.html


sgugger:
Thanks!

PR #31, opened 2019-12-21
batchsize error

holyfiddlex:
Just an added an error when batch size is bigger than dataset. A bit more helpful for debugging. Tested with:

    from fastai2.basics import *
    from fastai2.vision.all import *
    from fastai2.callback.all import *
    data = DataBlock(blocks=(ImageBlock, CategoryBlock), 
                 get_items=get_image_files, 
                 splitter=RandomSplitter(),
                 get_y=parent_label)
    path = untar_data(URLs.DOGS)

    bunch = data.databunch(path/"sample", item_tfms=RandomResizedCrop(460), bs=32, batch_tfms=[*aug_transforms(size=299, max_warp=0), Normalize.from_stats(*imagenet_stats)])

    bunch.show_batch()

### DIFF
--- a/fastai2/data/load.py
+++ b/fastai2/data/load.py
@@ -126,7 +126,7 @@ class DataLoader(GetAttr):
     def create_batch(self, b): return (fa_collate,fa_convert)[self.prebatched](b)
     def do_batch(self, b): return self.retain(self.create_batch(self.before_batch(b)), b)
     def one_batch(self):
-        if self.n < self.bs: raise ValueError(f'Batch size is {self.bs} but dataset only contains {self.n} items')
+        if self.n == 0: raise ValueError(f'Dataset does not contain any items')
         with self.fake_l.no_multiproc(): res = first(self)
         if hasattr(self, 'it'): delattr(self, 'it')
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks but this will break other things, as each time we use the command `Learn.predict`, we create a dataloader with less elements than the batch size. Or a language model dataloader could have only one very big text., so less than the batch size, but still have batches.

The error should be when the length of the dataloader is zero, do you mind adapting your PR in that sense?

holyfiddlex:
That makes sense, I changed the error but still think there should at least be a warning in the docs. DataBlock.databunch doesn't mention having a batch size parameter in the docs, only found it in the function signature.

Image of what I mean bellow, would be right after the tests in [DataBlock.databunch](https://dev.fast.ai/data.block#DataBlock.databunch). Tell me what you think to push these changes or you can just merge it the way it is now.
Screenshot_20191222_132941

sgugger:
Thanks for the change. The methods should have the kwargs in the doc, it's a bug (I think I know where it comes from and will fix).

PR #36, opened 2020-01-03
Update predictions callback options

lvaleriu:
After training a segmentation model i'd like to get the prediction losses on both train and validation subset in order to identify bad/difficult annotated data. I added the with_targs, with_preds parameters so that we can specify to get only the losses. Otherwise, on a big dataset, everything goes to ram memory. My 128 gb of RAM are not enough for some datasets.

Usage example:

pred_data = learn.get_preds(dl=learn.dbunch.train_dl, with_input=False, with_loss=True, with_preds=False, with_targs=False, with_decoded=False, act=None)
losess = pred_data[2]
tl = losses.topk(k=100) # worst data


sgugger:
Thanks for your PR! There is already the `save_preds`, `save_targs` argument to save the predictions/targets while going through the dataset without going OOM (should be a Path object in both cases).
This is adding more complexity when there is already a way to do this. Getting only the losses could be done in a separate function if it's really necessary, so I'm closing this.

PR #45, opened 2020-01-13
fix layer_info and fp16 for multi-target multi-output

mnpinto:


fastai2/callback/hook.py
```diff
@@ -146,7 +146,7 @@ def layer_info(learn):
     def _track(m, i, o):
         return (m.__class__.__name__,)+total_params(m)+(apply(lambda x:x.shape, o),)
     layers = [m for m in flatten_model(learn.model)]
-    xb,_ = learn.dbunch.train_dl.one_batch()
+    xb,*_ = learn.dbunch.train_dl.one_batch()
```
sgugger: Batches are always a tuple (of maybe nested tuples) so this should not be necessary.

fastai2/callback/fp16.py
```diff
@@ -89,7 +89,12 @@ def begin_fit(self):
         if self.dynamic: self.count = 0
 
     def begin_batch(self): self.learn.xb = to_half(self.xb)
-    def after_pred(self):  self.learn.pred = self.pred.float()
+
+    def after_pred(self):
+        if isinstance(self.learn.pred, list):
```
sgugger: There is the to_float method to do this normally.

fastai2/callback/hook.py
mnpinto: If the tuple len > 2 it gives this error:
image

fastai2/callback/fp16.py
mnpinto: Thanks, I will update it

fastai2/callback/hook.py
sgugger: Oh, it's actually something different, we need to cut the tuple at the right place. Let's merge this part as is and I'll follow up to add that part.

sgugger:
Thanks! You can check on the commit I referenced the correct code for grabbing the right xb.

PR #62, opened 2020-01-31
Added mention of nbdev to installation docs

drscotthawley:
I auto-generated `README.md` by editing `nbs/index.ipynb` and running `nbdev_build_docs`.
This process seems to have removed the header comment about the README file being auto-generated. 
Are we supposed to manually add the header after running `nbdev_build_docs`?
I'm happy to do more work before this PR gets merged.


sgugger:
I would reformulate. fastai v2 does not depend on `nbdev` as we discussed on the forum. If you want to play around in the notebooks and make changes/contribute then you need to install it, but if you just want to use the library, you don't need to add it.

sgugger:
Closing this PR as it's been a while. I added a different mention of nbdev in the index/README.

PR #63, opened 2020-01-31
feat: allow different number of input channels

borisdayma:
This PR let us change the number of input channels in loaded architectures, whether pretrained or not:

* I added the function `_get_first_layer` to make it easier to get/edit the conv (it has a different reference in alexnet or resnet for example). It is a homemade function but it worked with several different types of modes I tested.
* It works both with unet and cnn.
* I followed [Jeremy's comment](https://forums.fast.ai/t/fastai-v2-vision/53528/230) for pretrained arch:
  * with 1 channel, we take the sum of weights and biases
  * with 2 channels, we keep only the first 2 channels weights & biases and add 50%
  * with more than 3 channels, we set the additional channels weights & biases to 0
* bias is directly copied from previous architecture

I also added some tests (I needed to change the number of input layers to 3 in the `tst` model as per loaded architectures).


sgugger:
Thanks a lot for the PR. Can we keep the block that sets the weights of the first layer (the test in if `n_in != 3` in a separate private function? We like to have functions as short as possible with one purpose :) I can do it in a follow-up commit if you don't have time for this.

borisdayma:
I'll be happy if you can do it the way you typically like to organize the code.
Also since there are so many cases you may want to add other tests so that we have n_in=1, n_in=2, n_in=3 & n_in=4. I just didn't want to have too many tests but it may be useful in case of future commits.

borisdayma:
Should we also add an `assert` (after the one that verifies it's a Conv2d) to check that the number of input channels of loaded model is 3?

borisdayma:
Just to let you know in case you have not started working on it that I'm updating it

borisdayma:
Everything should be good to go!

sgugger:
Looking good to me, thanks a lot!

PR #64, opened 2020-01-31
fix learn.summary with multiple inputs

cwza:
I found this bug when I tried to print the summary of my seq2seq model. It would throw this error:  `<lambda>() takes 1 positional argument but 2 were given` when calling `learn.summary()` on `mutiple inputs` data.
Dig into the code and I think I find the problem, the little star is on the wrong place.
Fix it and also add some code to test layer_info on multiple inputs.
Thanks.


sgugger:
Thanks! I'll just hide the test form the docs in a follow-up commit (with the flag #hide) as it's pretty big (but it's good to have tests of the behavior you fixed, so thanks a lot for that!)

Issue #60, opened 2020-01-29
WandbCallback does not work


jakubLangr:
I think this is a function of the latest changes (that I generally massively support), but just thought I'd point out this issue. I will investigate it tomorrow. To see if I can fix it.

On the line in the camvid notebook:
`learn.fit_one_cycle(10, slice(lr), pct_start=0.9, wd=1e-2)` 
If we add (incl. appropriate imports & setup).
`learn.fit_one_cycle(10, slice(lr), pct_start=0.9, wd=1e-2, cbs=WandbCallback())`
It raises this issue:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/daisy-gan/venv/lib/python3.6/site-packages/fastai2/learner.py in fit(self, n_epoch, lr, wd, cbs, reset_opt)
    284             try:
--> 285                 self._do_begin_fit(n_epoch)
    286                 for epoch in range(n_epoch):

~/daisy-gan/venv/lib/python3.6/site-packages/fastai2/learner.py in _do_begin_fit(self, n_epoch)
    258     def _do_begin_fit(self, n_epoch):
--> 259         self.n_epoch,self.loss = n_epoch,tensor(0.);         self('begin_fit')
    260 

~/daisy-gan/venv/lib/python3.6/site-packages/fastai2/learner.py in __call__(self, event_name)
    222 
--> 223     def __call__(self, event_name): L(event_name).map(self._call_one)
    224     def _call_one(self, event_name):

~/daisy-gan/venv/lib/python3.6/site-packages/fastcore/foundation.py in map(self, f, *args, **kwargs)
    361              else f.__getitem__)
--> 362         return self._new(map(g, self))
    363 

~/daisy-gan/venv/lib/python3.6/site-packages/fastcore/foundation.py in _new(self, items, *args, **kwargs)
    314     def _xtra(self): return None
--> 315     def _new(self, items, *args, **kwargs): return type(self)(items, *args, use_list=None, **kwargs)
    316     def __getitem__(self, idx): return self._get(idx) if is_indexer(idx) else L(self._get(idx), use_list=None)

~/daisy-gan/venv/lib/python3.6/site-packages/fastcore/foundation.py in __call__(cls, x, *args, **kwargs)
     40 
---> 41         res = super().__call__(*((x,) + args), **kwargs)
     42         res._newchk = 0

[...]

PR #66, opened 2020-01-31
fix for issue #60

jakubLangr:
This solved the issue #60 for me. Was unable to pass tests locally as I have a Mac with non-CUDA pytorch.


sgugger:
This doesn't copy over the pipelines of transforms of the validation `DataLoader`. The correct fix is `dbunch.valid.new(...)` (without the `_dl`). Could you update your PR? Thanks!

jakubLangr:
Ok updated @sgugger, didn't reallise that thanks!

sgugger:
`dbunch` was for a a general `DataBunch`, here it's accessed through `self`, I just edited directly.
Thanks again!

PR #68, opened 2020-02-01
add warn if set attr in callback that also in learn

takotab:
https://forums.fast.ai/t/fastai-v2-callbacks-learner-optimizer/53519/17?u=takotab

### DIFF
--- a/fastai2/learner.py
+++ b/fastai2/learner.py
@@ -24,6 +24,11 @@ class Callback(GetAttr):
                (self.run_valid and not getattr(self, 'training', False)))
         if self.run and _run: getattr(self, event_name, noop)()
 
+    def __setattr__(self, name, value):
+        if hasattr(self.learn,name):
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks for your PR. First, please remember to run `nbdev_install_git_hooks` before pushing your PR as there is some metadata in the notebook you pushed that will create merge conflicts.
Then I'm not too sure about this warning as there are some situations where the callbacks and the learner both have attributes of the same name in the library and it would trigger this. Maybe we should add a debug mode and it only prints this warning in debug mode?

takotab:
cleaned the notebook. 

I do not have the overview to know if it would trip to much of the time. Debug mode sound super interesting. But it would only make sense if there are much more occasions where you could give feedback.

sgugger:
After further discussion with Jeremy, we decided that the situations were a Callback has an attribute of the same name as the Learner should be fixed, and therefore this warning should always be there.
Thanks for the PR!

PR #69, opened 2020-02-03
better datablock invalid kwargs error message

lgvaz:
Raises `TypeError` with a msg when invalid arguments are passed to `DataBlock`


lgvaz:
I don't really understand why `fastai2/callback/wandb.py ` was changed.. I have not touched that file

sgugger:
Don't see why either as there is no change. Maybe a different os line ending...
Anyway, thanks for the PR, this is a good idea! And thanks for adding a test :)

PR #74, opened 2020-02-05
adds RatioResize

lgvaz:
I've been using this function for a while in my style transfer experiments.  

After training the model I want to resize my images but maintain the aspect ratio, the current `Resize` function does not directly allow that.  

I thought about adding this as a another mode on `Resize` but I think that would be confusing to newcomers, because this method does not guarantees that all images will have the same size to collate a batch.


sgugger:
Looks useful, thanks a lot for the PR!

PR #78, opened 2020-02-06
fix learn.summary for multiple outputs

cwza:
I found this bug when I tried to print the summary from model that includes nn.GRU(). It would throw error: `TypeError: unsupported format string passed to list.__format__ ` 

Dig into the code and I find that _print_shapes() will produce list of string for multiple outputs but this formated string f'{_print_shapes(sz, bs):<20}' in summary() needs a single string. 
Fix it by just let _print_shapes() return string and limit its length for display.
Also add a test that check it will not break when the model has multiple outputs.

When I tried to fix this bug, I found another issue. If the output batch size is not at first dimension like nn.GRU() by default, it will return a wrong size. No good idea for solving this.

### DIFF
--- a/fastai2/callback/hook.py
+++ b/fastai2/callback/hook.py
@@ -154,7 +154,7 @@ def layer_info(learn):
 # Cell
 def _print_shapes(o, bs):
     if isinstance(o, torch.Size): return ' x '.join([str(bs)] + [str(t) for t in o[1:]])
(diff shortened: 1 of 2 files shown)


sgugger:
fastai only supports batch size first in text (otherwise it's impossible to have all function compatible with vision and tabular). Thanks for the PR!

PR #79, opened 2020-02-06
fixed ActivationStats#every

pete88b:
tiny change to make ActivationStats#every work


sgugger:
Makes sense! Thanks for fixing and adding a test!

PR #81, opened 2020-02-07
Fixed `PILLOW_VERSION` import error with pillow v7.0

akashpalrecha:
`torchvision` import breaks with pillow's recent update to v7.0.0. 

[Release notes for pillow v7.0.0](https://pillow.readthedocs.io/en/stable/releasenotes/7.0.0.html) say: `PILLOW_VERSION` has been removed. Use `__version__` instead.

Issue: `torchvision` tries to import `PILLOW_VERSION` which causes problems.

Fixed this by constraining pillow version to 6.2.1 in the `environment.yml` file. (`conda` does not yet have 6.2.2)

This is a workaround until a new version of `torchvision` is officially released.

More info here: 
https://github.com/pytorch/vision/issues/1712
https://github.com/pytorch/vision/issues/1712#issuecomment-570535277


sgugger:
No, we do not want to constrain pillow. There is a proper version of torchvision that deals with the import problem, if we did a constrain, it would be on this.

akashpalrecha:
I see. Just saw torchvision 0.5.0 deals with this already. (https://github.com/pytorch/vision/releases)

My bad. Should’ve checked properly.

PR #82, opened 2020-02-08
update Hook to optionally pass self to hook_func?

pete88b:
Can we write less code and/or gain any flexibility if we use 06_cuda_cnn_hooks_init.ipynb style hook functions? https://gist.github.com/pete88b/6e773b8c280f6e2db3500872caaf6d86


sgugger:
Mmmm, thanks for the PR but we designed func_kwargs to do exactly that (be able to pass a function at init and don't worry about self). Could you make a new PR using this instead? Look at the code of our `DataLoader` in data.laod for an example.

pete88b:
> we designed func_kwargs to do exactly that

Hi Sylvain,
Really sorry if I'm being thick but ... I can't find anything in foundation that does this.
The change I'm suggesting is to allow hook functions with either:
- 3 args (model, input, output) or
- 4 args (model, input, output, stored)
so that we don't have to use HookCallback classes in order to store things.
So .. as a fastai user, I can use 3 arg hook functions and HookCallback exactly as we do today - but ... if I want to access `stored` from a hook function, I can just add another argument.
I'll try to write an LSUV gist - think it'll demo this better

sgugger:
Yes, if you use @func_kwargs, like in the DataLoader, you can pass a function that takes 3 args and will be converted in a proper method. It makes no sense to pass a function taht takes 4 args as self will not really work (in that case people should subclass HookCallback).

Issue #57, opened 2020-01-28
Distributed Training not working after the latest commits


hussam789:
Hey,
After pulling the latest commit, it appears that Distributed Training stopped working.
we are getting this error: 
```Expected object of backend CUDA but got backend CPU for argument #2 'target' in call to _thnn_nll_loss_forward```

we checked the device type of `self.pred` on after_pred() and it appears that the `self.pred` is on the `cpu` and not on `cuda`

the latest commit that distributed training worked is around January 13'th.
maybe it's related to this commit: https://github.com/fastai/fastai2/commit/2c947668a6defeced761ce37b69be27504d72e2a

sgugger:
Could you give the code that fails (particularly the data creation process)? Thanks.

sgugger:
Should be fixed by #85, please reopen with a reproducible example if the issue persists.

PR #85, opened 2020-02-10
fix distributed training after DataLoader change (device parameter)

jaidmin:
fixes the distributed training issue introduces with the change from a `Cuda` callback to a `device` parameter on DataLoader reported in #57 and [here](https://forums.fast.ai/t/distributed-training-fails/63102) in the forums.

The fix is simply passing the `device` parameter from the `DataLoader` to the `DistributedDL` when it is initialized.

Also fixes this error:
```
warn(f"You are setting an attribute ({name}) that also exists in the learner. Please be advised that you're not setting it in the learner but in the callback. Use `self.learn.{name}` if you would like to change it in the learner.")
/home/jaidmin/Software/Devel/forks/fastai2/fastai2/learner.py:30: UserWarning: You are setting an attribute (dl) that also exists in the learner. Please be advised that you're not setting it in the learner but in the callback. Use `self.learn.dl` if you would like to change it in the learner.
```
If this behavior is wanted, let me know and I'll change the PR

### DIFF
--- a/fastai2/distributed.py
+++ b/fastai2/distributed.py
@@ -78,7 +78,7 @@ class DistributedDL(TfmdDL):
     @classmethod
     def from_dl(cls, dl, rank, world_size, **kwargs):
         cur_kwargs = dict(num_workers=dl.fake_l.num_workers, pin_memory=dl.pin_memory, timeout=dl.timeout,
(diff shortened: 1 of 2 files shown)


sgugger:
Looking good, thanks!

PR #86, opened 2020-02-10
Lesson1: 'pat' var not defined in resnet50 section.

pcuenca:


sgugger:
Thanks for your PR! It should be just one `\` with the r before, I think.

pcuenca:
You are right, but I think that's because of JSON escaping the `\` character - it looks fine in the notebook. I just copied the same pattern used in the resnet34 section :)

sgugger:
Oh ok. Should be good then, thanks!

PR #88, opened 2020-02-11
add model.summary(*sample_inputs)

cwza:
Sometimes I want to print shape and parameter counts when I am building the model. However the summary function is now bind to learner object, I have to first construct many things(DataLoaders, Datasets, Transforms, loss_func ...) before using summary(). Not so convenient. So I try to patch summary() function to nn.Module and keep the learn.summary functionality but change its implementation to call model.summary().
Hope this can help.

Usage:
``` python
m = nn.Sequential(nn.Linear(1,50), nn.ReLU(), nn.BatchNorm1d(50), nn.Linear(50, 1))
sample_input = torch.randn((16, 1))
m.summary(sample_input) 
```

Output:
```
Sequential (Input shape: ['16 x 1'])
================================================================
Layer (type)         Output Shape         Param #    Trainable 
================================================================
Linear               16 x 50              100        True     
________________________________________________________________
ReLU                 16 x 50              0          False     
________________________________________________________________
BatchNorm1d          16 x 50              100        True      
________________________________________________________________
Linear               16 x 1               51         True      
________________________________________________________________

Total params: 251
Total trainable params: 251
Total non-trainable params: 0
```


sgugger:
It's looking good and useful! I have some minor style issues but I'll push them in a separate commit after merging. Thanks!

PR #92, opened 2020-02-12
adds with_input option to learn.predict

lgvaz:
It's super useful to be able to return the input, I'll give my example:

I'm training keypoint regression on cat faces, after I trained my model I want to be able to use `.predict` to get my cat and the points back. The predict points will be relative to whatever transforms I used while training (e.g. Resize), they won't match with my original image.

It would be great if the predicted points matched with the original image instead of the transformed image, but I don't think that regards this PR. (And I'm also not sure how to do that)

Not my proudest work, but I tried to change little as possible to don't mess anything up. I have my git hooks set (same as I used in other PRs), I don't know why some other cells of the notebook changed.


fastai2/learner.py
```diff
@@ -328,12 +328,14 @@ def get_preds(self, ds_idx=1, dl=None, with_input=False, with_decoded=False, wit
...
-        dl = self.dls.test_dl([item], rm_type_tfms=rm_type_tfms)
-        inp,preds,_,dec_preds = self.get_preds(dl=dl, with_input=True, with_decoded=True)
-        i = getattr(self.dls, 'n_inp', -1)
-        full_dec = self.dls.decode_batch((*tuplify(inp),*tuplify(dec_preds)))[0][i:]
-        return detuplify(full_dec),dec_preds[0],preds[0]
+    def predict(learn, item, rm_type_tfms=None, with_input=False):
+        dl = learn.dls.test_dl([item], rm_type_tfms=rm_type_tfms)
+        inp,preds,_,dec_preds = learn.get_preds(dl=dl, with_input=True, with_decoded=True)
+        i = getattr(learn.dls, 'n_inp', -1) - 1
+        dec_inp, full_dec = map(detuplify, learn.dls.decode_batch((*tuplify(inp),*tuplify(dec_preds)))[0][i:])
```
sgugger: Putting the -1 on i doesn't necessarily give you the input (your model could have several inputs, like in tabular). i is the index to split between input and targets.
Those two lines should be rewritten with 
- no -1 for i
- dec_inp being everything up to i (not included)
- full_dec being everything from i and after (and also be renamed dec_targ maybe?)

sgugger:
Thanks for your PR! I think there is something wrong with your code however, so it needs a bit of change.

lgvaz:
Wait, I just saw something very wrong in this last commit. Fixing rn

lgvaz:
Cool, looks better?

I would add a test for multiple inputs, but I currently am not sure how to do that

sgugger:
Looking good thanks!

PR #93, opened 2020-02-12
adds PILImage.create from TensorImage

lgvaz:
It was the third project I needed to create a PILImage from a TensorImage, so I decided added into the lib =)


sgugger:
That's very useful, thanks!

PR #94, opened 2020-02-13
fix AWD_QRNN when bidir=True

cwza:
Bug happened when I called `AWD_QRNN` with `bidir=True`, it would cause this error: `"Can't save the previous X with bidirectional."`

I think the problem is AWD_QRNN always use save_prev_x=True to call QRNN, no matter bidir is what.

Fix it by assign flip of bidir to save_prev_x and add 2 simple tests.


sgugger:
Seems very reasonable. Thanks!

PR #101, opened 2020-02-14
integration with neptune

kamil-kaczmarek:
Hi,
Here is Kamil. Together with @jakubczakon, we created integration of neptune with fastai2. It works just like any other Callback and idea is similar to tensorboard callback, that is we collect logs during training.

Please let me know what do you think about it.

Best,
Kamil


sgugger:
It's looking good, thanks a lot! Just remember to run `nbdev_install_git_hooks` to have your notebook automatically cleaned at commit time. I'll merge and fix the problem here.

kamil-kaczmarek:
hey @sgugger thanks! I'll remember about `nbdev_install_git_hooks`.

PR #105, opened 2020-02-15
Upgrade torchvision: PIL version Issue

krishnakalyan3:
https://github.com/pytorch/vision/issues/1712

Recently PIL 7 removed `PILLOW_VERSION`. This affects `torchvision: 0.4.2` as described in the pytorch issue.

Alternative upgrade to `torchvision>=0.5.0`

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,13 +6,13 @@ channels:
 dependencies:
 - jupyter
 - pytorch>=1.2.0,<1.4.0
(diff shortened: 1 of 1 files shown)


sgugger:
We can't do this for now as there is a bug in torchvision 0.5.0 with current fastai v2. We will constrain to that version as soon as that bug is fixed.

Issue #87, opened 2020-02-11
Memory management with WandbCallback after finishing an epoch


jakubLangr:
I am stuck on this for several days so I thought I’d post it here.

I find there is something odd with the memory management after finishing an epoch. I tried to run Fastai U-Net on the Synthia dataset and this is an error I get despite having a batchsize 1 with a crop to 360 on an 11GB K80. I must be missing something, because camvid runs fine: (also does not seem to happen if I only load 10 examples to the dataset)
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-18-637f3e7802b9> in <module>()
----> 1 learn.fit_one_cycle(10,slice(1e-6,1e-3), cbs=WandbCallback())

~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/fastai2/callback/schedule.py in fit_one_cycle(self, n_epoch, lr_max, div, div_final, pct_start, wd, moms, cbs, reset_opt)
     88     scheds = {'lr': combined_cos(pct_start, lr_max/div, lr_max, lr_max/div_final),
     89               'mom': combined_cos(pct_start, *(self.moms if moms is None else moms))}
---> 90     self.fit(n_epoch, cbs=ParamScheduler(scheds)+L(cbs), reset_opt=reset_opt, wd=wd)
     91 
     92 # Cell

~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/fastai2/learner.py in fit(self, n_epoch, lr, wd, cbs, reset_opt)
    290                         self._do_epoch_validate()
    291                     except CancelEpochException:   self('after_cancel_epoch')
--> 292                     finally:                       self('after_epoch')
    293 
    294             except CancelFitException:             self('after_cancel_fit')

~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/fastai2/learner.py in __call__(self, event_name)
    221     def ordered_cbs(self, cb_func:str): return [cb for cb in sort_by_run(self.cbs) if hasattr(cb, cb_func)]
    222 
--> 223     def __call__(self, event_name): L(event_name).map(self._call_one)
    224     def _call_one(self, event_name):
    225         assert hasattr(event, event_name)

~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/fastcore/foundation.py in map(self, f, *args, **kwargs)
    360              else f.format if isinstance(f,str)
    361              else f.__getitem__)
[...]

PR #106, opened 2020-02-15
WandbCallback: fix issues

### LINKED ISSUE #70: WandbCallback does not work with TfmdLists
When using `TfmdLists`, WandbCallback returns an error at [this line](https://github.com/fastai/fastai2/blob/master/fastai2/callback/wandb.py#L47).

The reason is that `self.dls.valid_ds` is `TfmdLists` so has no `tls` attribute (vs `Datasets` which is expected here).

At this step, we are trying to create a dataloader with specified idxs for future inference.

On another note, a few other issues are related to this callback:
* we sometimes run out of CUDA memory so maybe should do the prediction on CPU (since it won't happen too often).
* `valid_dl` may have a different batch size (maybe we should get the one from the fit loop) than the number of predictions (default to 36) so we would have to make sure we either run it a few times (if bs < n_preds) or check we have all results.

borisdayma:
This PR:

* fixes #70 
* allow logging any dict for more flexibility -> we can improve `wandb_process` later
  * log `best/worst predictions`
  * separate `inputs`, `labels`, `predictions` to display them separately or together. This also let us display evolution of predictions along time (this is what I'm currently doing)
  * allow custom `wandb_process` for users to log whatever they want

### DIFF
--- a/fastai2/callback/wandb.py
+++ b/fastai2/callback/wandb.py
@@ -42,10 +42,8 @@ class WandbCallback(Callback):
             wandbRandom = random.Random(self.seed)  # For repeatability
             self.n_preds = min(self.n_preds, len(self.dls.valid_ds))
             idxs = wandbRandom.sample(range(len(self.dls.valid_ds)), self.n_preds)
-
-            items = [self.dls.valid_ds.items[i] for i in idxs]
-            test_tls = [tl._new(items, split_idx=1) for tl in self.dls.valid_ds.tls]
-            self.valid_dl = self.dls.valid.new(Datasets(tls=test_tls), bs=self.n_preds)
+            test_items = [self.dls.valid_ds.items[i] for i in idxs]
+            self.valid_dl = self.dls.test_dl(test_items)
 
     def after_batch(self):
         "Log hyper-parameters and training loss"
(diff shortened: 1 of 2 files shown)


borisdayma:
I'm now also going to look at #87

borisdayma:
@sgugger `get_preds` seems to solve CUDA memory issues.

I get `its` and `outs` through:

```python
inp,_,outs,dec_preds=self.learn.get_preds(dl=self.valid_dl, with_input=True, with_decoded=True)
outs = self.dls.decode_batch((*tuplify(inp),*tuplify(dec_preds)), max_n=None)
its = self.dls.decode_batch((*tuplify(inp),*tuplify(outs)), max_n=None)
```

However I still have a few issues.

`get_preds` makes the progress bar disappear.

Also for the type dispatch to work, I still need to extract x & y from `its` but it seems redundant as I have the impression that `its` (or `samples` as named in `wandb_process`) is just made of `x` and `y`.

sgugger:
Yes, I was pretty sure something like `get_preds` would fix the issue since it stores predictions on the CPU. I need to refactor the part that does custom preds and make that easily accessible in callbacks, since it's used in Tensorboard or GANs as well to display predictions. I hope to have some time to work on this this week.

sgugger:
In the meantime, this looks great to fix #70 so let's merge :)

borisdayma:
Ok, the other advantage is that we also get all the predictions we want (instead of only one batch).

Let me know if I can be of any help.

PR #109, opened 2020-02-18
test(wandb): add test

borisdayma:
This PR adds a test for WandbCallback


sgugger:
Thanks for the test! Will try to fix the problem when I have time.

PR #113, opened 2020-02-18
fix: DataBlock respect dl_type kwarg

davecoates:
Looks like this was inadvertently removed in https://github.com/fastai/fastai2/commit/dcc7e1bfa7170700143127e56d8dae8e97ef8177#diff-3fec432759d383a373137d69c9ac4c72L58?


sgugger:
Indeed, thanks a lot for putting it back!

PR #114, opened 2020-02-18
added delegate kwargs of subplot to get_grid

takotab:
I wanted this for the sharex and sharey options but with kwargs it makes it more flexable


sgugger:
Thanks!

PR #116, opened 2020-02-19
fix bbox coordinates convention

manycoding:
I noticed that the convention [changed](https://github.com/fastai/fastai2/blob/master/fastai2/vision/core.py#L175) from v1 to x1,y1,x2,y2.


sgugger:
I'll add a little bit more prose but this is right. Thanks for catching it!

PR #118, opened 2020-02-20
PathLabeller

nareshr8:
Add PathLabeller as a labelling function


sgugger:
Thanks for your PR, but this duplicates the RegexLabeller. I don't understand what new functionality this is adding.

nareshr8:
I had to get the label from grandparent. Getting Grandparent through Regex was hard for me. So I modified RegexLabeller a bit to accommodate the same and shared. If it can be achieved through RegexLabeller feel free to cancel the pull request.

sgugger:
Let me check, but you should always be able to pass a list of function to `get_y` in the data block API (if that's not the case I will fix). So you should be able to say `get_y = [grand_parent, RegexLabeller(...)]` with grand_parent being a function that returns the grand parent name. I think this a bit more modular instead of having all the pieces in one big block.

I'm closing this PR and checking that the above works (and if it does not, will fix this morning :) )

sgugger:
Just updated the example for the pets dataset in 50_datablock_examples to use a list in get_y, it works perfectly ;)

PR #119, opened 2020-02-20
Show rules to Tokenizer calss in .from_df

morganmcg1:
When calling Tokenizer.from_df, rules=rules lies inside the call to get_tokenizer and so the Tokenizer class defaults to the default fastai rules as the class cannot see any fules added in .from_df()


morganmcg1:
https://github.com/fastai/fastai2/commit/6ec16e6ded84a52d73052a0d17adb7780dbd813d adds `rule=rules` to the correct place in `from_folder` but not in `from_df`, resulting in fastai special tokens being tokenized by a custom tokenizer (like the hugginface tokenizers)

Tested the above change by passing `rules=[]` to `Tokenizer.from_df` and the fastai special tokens did not appear

sgugger:
Ah thanks!

PR #120, opened 2020-02-20
add sklearn traintestsplitter

lvaleriu:


sgugger:
I am now super familiar with this function from scikit-learn. Could you explain what it adds on top of our RandomSplitter?

lvaleriu:
It adds the stratify argument that allows splitting uniformely according to the label distribution. Should I add this argument into the notebook too?

sgugger:
Yes, it's best if the user can see all the arguments at their disposal. Don't know if @delegates is useful here or if scikit-learn hides things in kwargs.

lvaleriu:
I’ll add the most common arguments then (like random_state, stratify, test_size)

sgugger:
Can you please put the second line of documentation in a markdown cell of the notebook? We like the docstrings to be one line only, and all additional prose to be in the notebook itself.

lvaleriu:
I had a bad feeling about it too and was hesitating to ask how you do when the docstrings are long :)

sgugger:
Thanks a lot, this is looking good!

PR #129, opened 2020-02-28
Allow TabularModel to bypass BatchNorm1d for continuous inputs

ourownstory:
Added an additional argument `bn_cont` to the `TabularModel` to allow to bypass `BatchNorm1d` on the continuous input variables `x_cont`. 
Also added the argument `bn_cont` to the wrapping `tabular_learner` function.

Reason: For some models, the direct relationship between input values and model weights matters. Thus,  `BatchNorm1d` needs to be bypassable. 

Example: Regression or Auto-Regression.


fastai2/tabular/model.py
```diff
@@ -28,12 +28,13 @@ def get_emb_sz(to, sz_dict=None):
...
     "Basic model for tabular data."
     def __init__(self, emb_szs, n_cont, out_sz, layers, ps=None, embed_p=0.,
-                 y_range=None, use_bn=True, bn_final=False):
+                 y_range=None, use_bn=True, bn_final=False, bn_cont=True):
         ps = ifnone(ps, [0]*len(layers))
         if not is_listy(ps): ps = [ps]*len(layers)
         self.embeds = nn.ModuleList([Embedding(ni, nf) for ni,nf in emb_szs])
         self.emb_drop = nn.Dropout(embed_p)
-        self.bn_cont = nn.BatchNorm1d(n_cont)
+        self.bn_cont = None
```
sgugger: This test can be written on one line `self.bn_cont = nn.BatchNorm1d(n_cont) if bn_cont else None`.

ourownstory: Updated.

sgugger:
LGTM, thanks a lot!

PR #131, opened 2020-02-28
Refactor xresnet

jaidmin:
Here the pull request refactoring `XResNet`. Some other files are affected, I am not sure why, I pulled straight from master and ran `nbdev_clean_nbs` and `nbdev_build_lib`.


sgugger:
Must be some weird merge that makes those appear as diff. Thanks a lot for the PR!

PR #134, opened 2020-03-02
Allow user to write custom Regularizer

ourownstory:
Added an optional `reg_func` argument to the `Learner` class.
When training, the `reg_func` gets called with `self.model` as an argument.
The returned value gets added to `self.loss`.

Motivation: Allow users to write custom regularizers in a simple manner.


sgugger:
No, we don't want to add this directly to the training loop. This should be done in a callback instead (see RNNRegularizer for instance).

PR #137, opened 2020-03-03
Added SubsetSplitter, similar to v1

foobar8675:
I was looking for the SubsetSplitter in v2, but did not find it. So this is my attempt to add one that would be compatible with v2


sgugger:
It looks like your PR is using an old version of nbdev, and your export has changed a lot of things. Could you upgrade nbdev and amend it? Or if it's too complicated, could you change it to have only the modifications in the notebook and I'll do the conversion after merging?

foobar8675:
Sure, I'll do the upgrade today or tomorrow and try again.

foobar8675:
I ran a `pip install nbdev` and cut a fresh branch off master but after doing a `nbdev_build_lib` I'm still seeing 5 files changed

```
	modified:   fastai2/_nbdev.py
	modified:   fastai2/callback/tensorboard.py
	modified:   fastai2/callback/wandb.py
	modified:   fastai2/data/transforms.py
	modified:   fastai2/vision/learner.py
```

I'm going to push my new branch with only the nbs changes as per your second suggestion and close this PR.

sgugger:
pip install nbdev is not the most recent version of nbdev, you need a developer install for the latest unreleased change. But like I said, just the notebook changes is enough. Thanks!

PR #140, opened 2020-03-03
Adding SubsetSplitter

foobar8675:
Adding SubsetSplitter to nbs


sgugger:
Thanks!

sgugger:
Mmm studying this before exporting, but it's not a splitter since it does not take a list of items, but thins that are already split. Yet you have an assert for the sum of the sizes.
I rewrote it a bit to actually take splits from a list of items and adapted the test.

PR #141, opened 2020-03-04
fix AWD_LSTM not contiguous bug

cwza:
Error: `rnn: hx is not contiguous`

This error happens when 
1. bidir=True 
2. bs decrease
3. run on gpu

The 1 and 2 cause `self.hidden[l][0][:,:bs]` be not contiguous, and with 3 the error happens.

Fix it by adding contiguous() after self.hidden[l][0][:,:bs] and add a test that can reproduce this error with cuda flag.


sgugger:
Thanks a lot for fixing!

PR #143, opened 2020-03-05
adds support for multiple folders on GrandparentSplitter

lgvaz:
Adds the possibility to pass multiple names to `train_name` or `valid_name`, why is this useful?

Let's say my data is separated in `train/valid/test`, all datasets are labeled. What I want to do is use `train` and `valid` to find the best model/hyperameters and then `train+valid` and `test` at the end to calculate my final metrics.

For testing this new functionality I had to create another variable called `fnames2` to add the "test" folder. I cannot just append to `fnames` or else the tests starts failing for `FuncSplitter`, since it will consider the "test" items as training.

I have to say I find the way `FuncSplitter` works a bit counter intuitive, every time the function returns `False` the item just goes to the training set, it's very easy to mistakenly add items to the training set, as it happens with the example above. Do you think it's a good idea to do something like: "Return 0 for training, 1 for valid, and -1 to ignore"?

Some other unrelated errors popped when I ran all the notebook, I was confused if I should leave or not the errors. I left them here just to be on the safe side


lgvaz:
For some context, below is the strack trace for one of the unrelated errors I got in the nb.
I don't know if there is something wrong on my side.
I can easily strip out the errors if think it's a better idea.
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-61-b42e06313315> in <module>
      3 test_eq(tds[3][0], ())
      4 test_eq(cat.vocab, ['a', 'b', 'c'])
----> 5 test_eq(cat(['a', 'c']), tensor([0,2]))
      6 test_eq(cat([]), tensor([]))
      7 test_eq(cat.decode([1]), ['b'])

~/libs/fastcore/fastcore/test.py in test_eq(a, b)
     30 def test_eq(a,b):
     31     "`test` that `a==b`"
---> 32     test(a,b,equals, '==')
     33 
     34 # Cell

~/libs/fastcore/fastcore/test.py in test(a, b, cmp, cname)
     20     "`assert` that `cmp(a,b)`; display inputs and `cname or cmp.__name__` if it fails"
     21     if cname is None: cname=cmp.__name__
---> 22     assert cmp(a,b),f"{cname}:\n{a}\n{b}"
     23 
     24 # Cell

~/libs/fastcore/fastcore/imports.py in equals(a, b)
     64     if one_is_instance(a,b,type): return a==b
     65     if hasattr(a, '__array_eq__'): return a.__array_eq__(b)
---> 66     if hasattr(b, '__array_eq__'): return b.__array_eq__(a)
     67     cmp = (np.array_equal if one_is_instance(a, b, ndarray       ) else
     68            operator.eq    if one_is_instance(a, b, (str,dict,set)) else

~/git/fastai2/fastai2/torch_core.py in __array_eq__(self, b)
     94 @patch
     95 def __array_eq__(self:Tensor,b):
---> 96     return torch.equal(self,b) if self.dim() else self==b
     97 
     98 # Cell

TypeError: equal(): argument 'other' (position 2) must be Tensor, not tuple
```

sgugger:
Thanks for your PR! There are a few things to fix before we can merge it.
Please use four spaces for indentation and not two. Also, since you are changing the function `_grandparent_idxs`, remove the old version since it doesn't make sense to have it.
You have weird things exported, make sure you upgrade your nbdev to the last version before running the command to build the library.

lgvaz:
Hi, I updated nbdev and built the library again, but it seems does weird things exported still there? 
Like this one:
```
__all__ = ['TensorBoardCallback', 'tensorboard_log']
```

I don't understand why those files changed since I only touched the `data.transforms.ipynb` file.

Do you think it's a better idea to start from a clean state and do the modifications to `_grandparent_idxs` again?

sgugger:
I can fix it in a separate commit, the build works fine on my side.

PR #144, opened 2020-03-05
Improved DICOM Support

moritzschwyzer:
I added the function `get_dicom_files` analog to `get_image_files` in `medical.imaging` to easily access DICOM files in a given folder.
Additionally, I modified the `scaled_px` function in `medical.imaging` so that it doesn't use the Rescale Slope and Rescale Intercept DICOM tags if an x-ray DICOM file is used, because Rescale Slope and Rescale Intercept are only needed for CT DICOM datasets.
Lastly, I added the `force=False` argument to `Path.dcmread`. If DICOM headers have missing data, `force` can be set to `True` so that the DICOM data can be loaded anyways.


fastai2/medical/imaging.py
```diff
@@ -34,7 +41,10 @@ def pixels(self:DcmDataset):
 def scaled_px(self:DcmDataset):
     "`pixels` scaled by `RescaleSlope` and `RescaleIntercept`"
     img = self.pixels
-    return img*self.RescaleSlope + self.RescaleIntercept
+    if self.Modality == "CR":
```
sgugger: This test can fit in one line: 
`return img if bla else ...`
Would you mind adjusting your PR with this?

sgugger:
Thanks, this looks useful! Could you just address the one comment I have before we merge this?

moritzschwyzer:
Thanks for the input! I just rewrote the return statement in one line.

sgugger:
Thanks!
You have some weird exports, probably due to an old version of nbdev. Will rerun the export script to fix after merging this PR.

PR #152, opened 2020-03-07
Adding ValueMetric

ohmeow:
This subclass of Metric allows for adding metrics to Learner calculated elsewhere (e.g., Callbacks)


fastai2/learner.py
```diff
@@ -369,6 +369,18 @@ def accumulate(self, learn):
     @property
     def value(self): return self.val/(1-self.beta**self.count)
 
+# Cell
+class ValueMetric(Metric):
+    "Use to include a pre-calculated metric value (e.g., a metric calculated in a `Callback` and returned in `func`)"
+    def __init__(self, func, metric_name:str=None):
+        store_attr(self, 'func, metric_name')
```
sgugger: I think this can all fit in one line. Also note we have stopped using type annotations (it's pretty obvious `metric_name` should be a string here).

sgugger:
That seems useful, thanks for the PR! Just one little cosmetic comment if you don't mind addressing, and we can merge.
Thanks for adding a test!

ohmeow:
Sure. I'll resubmit in a bit.

BTW, I think this exists but not sure where, but is there a fastai coding
style doc available? Even better would be some kind of linter in nbdev.

sgugger:
The style guide is that there is no style guide ;) We sill abide by the style guide of v1, mostly our main thing is to try to limit vertical space whenever possible. So one-line functions or one-line tests when it's still readable.

PR #154, opened 2020-03-08
fix RocAucMulti metric

ohmeow:
fixed the metric and added a test to verify


sgugger:
Thanks a lot for fixing!

PR #157, opened 2020-03-09
Select partial quantity of data per epoch

borisdayma:
Allows selection of partial quantity of data per epoch.

At each epoch, `partial_n` items are selected randomly from the full dataset and are then replaced for potential selection in next epoch.

This let us have more feedback by processing validation set more frequently on large datasets.

It could be improved by allowing no replacement until full dataset has been processed.
Let me know if that can be of any interest and I'll update the PR.


sgugger:
Looking good, thanks!
Sidenote: I will fix in a follow-up commit but we prefer single double-quotes for the docstring.

PR #159, opened 2020-03-09
Adding OptimizeFBetaThreshCallback callback

ohmeow:
Allows users to find the optimal threshold for multi-label problems based on FBeta definition. Downstream metrics with a "thresh" attribute/argument will be modified by default so that their results will use the found threshold.


sgugger:
Thanks a lot for the PR. After discussing it with Jeremy, we agree it's a bit too niche to be in fastai v2. Not saying it's not useful so you should definitely package it in your own repo then add a link to it on the forum.

We plan to add some tutorials to package fastai submodules in the near future (for instance, medical will be a submodule and not inside fastai, or all the extensions for tensorboard, wandb or neptune) and you package this in something that would be fastai.something if you want.

ohmeow:
Understood.

I'll just hold on to this until you all flesh out how adding submodules
will work and what they should be called. I'm sure y'all don't want a
million of them.

Thanks

PR #160, opened 2020-03-11
Replace nn.Module by Module on layers.py

tcapelle:
Just doing so


sgugger:
I thought I had done it but I guess I forgot. Thanks!

PR #168, opened 2020-03-13
fixes to_concat with dict

lgvaz:
Before the fix the following error was thrown when calling `to_concat` with a list of dicts:

```python
AttributeError: 'list' object has no attribute 'keys'
```

I also added a very simple test, mainly to check if no error is thrown when using a dict.
To actually test the expected result we would need a recursive `test_eq`. Maybe we currently have something that does that, but I'm unaware of.


sgugger:
Thanks for fixing this!

PR #169, opened 2020-03-13
Fix Jaccard and JaccardMulti to use jaccard_score instead of jaccard_similarity_score

jaeeolma:
Current Jaccard and JaccardScore use jaccard_similarity_score which has been deprecated (and broken) since scikit-learn v 0.21( [scikit-learn docs](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.jaccard_similarity_score.html#sklearn.metrics.jaccard_similarity_score)).  Correct method is jaccard_score.


sgugger:
Thanks for the PR!
I will merge for the fix but be careful to update nbdev to the last version next time. You have a lot of diff showing some weird exported stuff that should not be there.

PR #171, opened 2020-03-13
Update metrics.py

miwojc:
Fixed typo (missing coma) in line 111


sgugger:
Addressed it too, but thanks!

Issue #148, opened 2020-03-05
Memory Usage by Preprocessors


muellerzr:
Opening up a different issue here. On Rossmann if I do not preprocess my categorical datatypes, the footprint goes up by 1gb. However if I preprocess by doing:

```python
for name in cat_vars:
  train_df[name] = train_df[name].astype('category')
to = TabularPandas(train_df, procs=[Categorify], cat_names=cat_vars, cont_names=cont_vars,
                   y_names=dep_var, block_y=TransformBlock(), splits=splits, inplace=True)
```
the footprint stays at ~300mb's (the same as it was after doing the forced `astype`. I'll try to take a look.

muellerzr:
For a better visualization:

Before converting:
```
              total        used        free      shared  buff/cache   available
Mem:          13022        1540        7518           0        3963       11624
Swap:             0           0           0
```
When doing the `for name etc`:
```
              total        used        free      shared  buff/cache   available
Mem:          13022        1817        7277           0        3928       12040
Swap:             0           0           0
```
(which stays the same after doing `TabularPandas`
Without doing so:
```
              total        used        free      shared  buff/cache   available
Mem:          13022        2829        6226           0        3966       11614
Swap:             0           0           0
```

muellerzr:
I did some more work into looking into this. I adjusted the title of the issue to reflect what I looked into. Instead I compared how the memory usage changed in regards to each preprocessor, and here is what I found:

| Name | N rows | Pre-memory | Post-Memory | Secondary  Post-Memory  usage | Added Memory | % increase |
|:-----------------------------:|:-------:|:----------:|:-----------:|:-----------------------------:|--------------|------------|
| All 3 with no cat pre-process | 844338 | 1.53gb | 3.17gb |  | 1.64gb | 107.19 |
|  | 2000000 | 3.98gb | 6.61gb |  | 2.63gb | 66 |
| All 3 with cat pre-process | 844338 | 1.53gb | 1.81gb | 1.85gb | 0.32gb | 20.9 |
|  | 2000000 | 3.98gb | 3.87gb | 6.28gb | 2.3gb | 57.78 |
||||||||
| Categorify no pre-process | 844338 | 1.53gb | 2.40gb |  | 0.87gb | 56.86 |
|  | 2000000 | 3.98gb | 4.17gb |  | 0.19gb | 4.7 |
| Categorify with pre-process | 844338 | 1.53gb | 1.81gb | 1.85g | 0.32gb | 20.9 |
[...]

PR #173, opened 2020-03-14
Reduce memory from TabularPandas

muellerzr:
As discussed [here](https://github.com/fastai/fastai2/issues/148), this was my first thought on it. Let me know if you were thinking of something a little bit different, or if I should change anything at all. (I also added a check to see if we already set the pandas chained option)


muellerzr:
Don't merge this quite yet please, unexpected bug :)

muellerzr:
Fixed the bug, testing on Rossmann footprint was reduced from 2.6gb to 2.15gb (Matching the experiments)

sgugger:
This looks good but you have quite a few weird exports. Do you have the latest `nbdev`?

muellerzr:
Good catch, I do not! Let me try again

muellerzr:
Looks as though that fixed it all (except metrics.py for some reason)

sgugger:
Better, thanks! Does this close #148?

muellerzr:
Yes it does!

sgugger:
The `reduce_cats` make the tests later in the book fail (I commented it out for now). The problem is that the categories should be decided on the training set (if there is one) and then used on the validation set (which is the whole point of `Categorify`)

muellerzr:
Got it, yea there was a bug pointed out earlier on the forums about this. I'll make the adjustment I think is needed and push a PR shortly

PR #174, opened 2020-03-14
Small fix to 10_tutorial.pets notebook

wdhorton:
There was still one reference to `Datasource`, so I changed it to use the new name `Datasets`.


sgugger:
Thanks a lot!

PR #175, opened 2020-03-15
Removed dependency with fastai v1 on Lesson2-sgd notebook

artste:
In order to run the notebook *without fastai v1 installed* and to show that this is plain PyTorch code and no magic is happening behind the scene, the dependency:
```python
from fastai.basics import *
```
Has been substituted with:
```python
# fastai v1 back compatibility
import matplotlib.pyplot as plt
import torch
import torch.nn as nn
import numpy as np

def tensor(*argv): return torch.tensor(argv)

# TEST
assert torch.all(tensor(1,2) == torch.tensor([1,2])), 'Simpler way to instance tensor'
```


sgugger:
This is great! Be careful to run `nbdev_install_git_hooks` however, before making any PR, as your notebook is not stripped of its metadata. I'll merge an clean up for this time.

artste:
Thnx!

PR #179, opened 2020-03-15
Fix "None"

muellerzr:
Needed to be NoneType, not "None" for pandas warning


muellerzr:
This *should* be the last mini-PR that I've discovered until I explore the `.setup()` (and if there's anything I can really even do there)

muellerzr:
Don't merge it quite yet though, in case I do let's use this as the one PR :)

muellerzr:
Okay, good to merge. I was actually going crazy and didn't realize that doing `df = df.append(df)` will essentially multiply your memory usage. Memory reduction looks much more sane now, and added overhead even with a 6.5gb file is only 8.1gb

sgugger:
Thanks!

PR #183, opened 2020-03-18
Fix categorical issue

muellerzr:
This should fix it, let me know otherwise if you had a different idea


sgugger:
No, this does not change the fact that categories are set on the whole dataframe instead of just the training part. Make sure all the notebook can run with your change, you should see the tests of Categorify fail.

muellerzr:
Got it, my apologies. Let me look back at it real quick

muellerzr:
Tests pass now :)

sgugger:
I think you can put all the conts to float, no?

muellerzr:
Yes, you are right. We wouldn't have the same issue there

sgugger:
Thanks!

muellerzr:
Thank you for helping me work through it :)

PR #186, opened 2020-03-19
changes split_idx default on RandomErasing

lgvaz:
I think `RandomErasing` should be applied to the validation set as default.

My use cases so far were:
* Image inpainting
* Classification (I want to see how good my model is with "corrupted images")

I totally understand if this PR does not get accept, since this is a very subjective change =)


sgugger:
Thanks for the PR. You should define your subclass for that problem. In general we do not want data augmentation applied to the validation set, so this change would surprise many users.
So I'll close this ;)

PR #187, opened 2020-03-20
add docs to data.external

amaarora:
Update docs for data external


amaarora:
This should have built docs for data.external but nbdev_build_docs doesn't work for me. Might need help to fix this PR.

amaarora:
Just added docs/data.external.html to the PR as well. :)

amaarora:
Please not that I haven't added an explaination for `MNIST_VAR_SIZE_TINY`, `FOOD` datasets as I wasn't sure what they are. I have tried to best guess the datasets available inside the URLs class and provide links to the research papers or datasets available on the internet. 

Have also added examples to the CONFIG class. 

Shown a useful example to download a google image using `download_url` function. And finally updated the docs for `untar_data`.

sgugger:
Thanks a lot!
Note that in general, we don't want contributors to build the docs themselves to avoid large diffs. I trigger the doc building after merging PRs that altered them.

PR #189, opened 2020-03-20
Fix last by_block typo

muellerzr:


muellerzr:
Just realized library didn't build, apologies, one moment!

muellerzr:
Actually, no. The last by_block was fixed in the .py but it wasn't updated in the notebook

sgugger:
Thanks a lot!

PR #191, opened 2020-03-21
add MACAQUES dataset

radekosmulski:
This is a small dataset containing recordings (saved as wav files) of macaque coo calls recorded from 8 individuals.

They come from this [paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4806230/), have been put in public domain and have been made available [here](https://datadryad.org/stash/dataset/doi:10.5061/dryad.7f4p9).

I downloaded the data and store it on GCP under my account for ease of use and speed of access.


sgugger:
Thanks a lot! 
For consistency with the rest of the datasets, I will copy it to our server then follow up with a commit to change the URL (maybe not today but sometime soon). Once that commit is passed, you'll safely be able to remove it from your GCP.

radekosmulski:
sounds great and makes perfect sense :) thx for the merge!

PR #192, opened 2020-03-22
create new hash for the macaque dataset

radekosmulski:
My apologies, I messed things up a little bit. I now zipped the directories containing the data in a way that makes sense and calculated a new hash for the macaque dataset.

If you already grabbed the zip file from my GCP, could you please grab the new one I uploaded?

sorry for the trouble!


sgugger:
I had not downloaded it yet, so all good :)

PR #193, opened 2020-03-22
bugfix-190 platform is Ubuntu18.04 

RogerS49:
bugfix-190


sgugger:
Thanks for your PR, but this has already been done in [this commit](https://github.com/fastai/fastai2/commit/93d6648f1b28be23e4021125cd4218ed85880633).

PR #194, opened 2020-03-22
Add gradient accumulation

borisdayma:
Support gradient accumulation through `GradientAccumulation` callback.


borisdayma:
Next I'll look how to make gan switchers compatible with this callback so they don't switch while we are just accumulating gradients.

borisdayma:
I think I'm going to add `self.loss = AvgLoss` and update it accordingly.

This way I can (if callback present):
* use `self.gradient_accumulate.count == 0` to add an iteration to `FixedGanSwitcher`
* same as previously + check `self.gradient_accumulate.loss.value` for `AdaptiveGanSwitcher`

sgugger:
I think for the specific use of gradient accumulation with GANs we can have another specific callback. Merging this for now, and I may reorganize the callbacks a bit after since it has little to do with tracker.

borisdayma:
Noted, I'll see if I can propose something.

For the moment I'm having some strange issues with gradient accumulation where my loss (and gradients) suddenly explode after a while.

image

In this example I accumulate 10 batches. It does not happen with 1 batch (ie no gradient accumulation) and happens later if I decrease number of batches accumulated or if I decrease learning rate (even by a lot), which means it's probably not the dataset.

I'm trying to see if there is an issue with this implementation but the gradient update and optimizer steps seem to be performed at the right moment and the individual losses `self.learn.loss` seem to remain bounded.

Something interesting is if I stop the training after it "explodes" and then restart, the model will learn again -> [run that restarts after 100 epoch](https://app.wandb.ai/borisd13/l2ab/runs/1dt862p7?workspace=user-borisd13)

I'm now investigating a few things:
1. is it due to the optimizer?
  * with `mom=0` on Adam it's still a problem
  * with SGD it seems to not explode (though it's difficult to get fast learning)
2. does it happen if I scale down the loss before calculating gradients? I initially thought that scaling down the loss at each gradient calculation or scaling down learning rate before weight update would be the same but I'm not so sure anymore.
3. looking at run where I restarted training and histograms of gradients and parameter values, it seems that the problem could be caused by the batch norm layers (bias in particular).

Anyway, I'm investigating further but let me know if you have any idea of what could be the issue.

sgugger:
What's your fir method? If you call a method with warm-up like fit_one_cycle, you should multiply the number of epochs by 10 to ensure the warmup is not too strong.

borisdayma:
I use fit_one_cycle.
Ok I'm gonna try a larger number of epochs. Right now I had 100.
Also my epochs are not full epochs as the dataset is really large so I use `partial_dataloaders` with each epoch = 1000 batches (this let me have more validation feedback).

It makes me realize that this could also be a failing point.

borisdayma:
> Also: scaling the loss is necessary IMO, since we usually average over batches. So you have gradients 10 times stronger than usual in your example.

I was thinking that if I divide my learning rate by 10 it would be equivalent.

But it is assuming that
image

It may be incorrect to assume that the gradients are proportional to the loss.

sgugger:
Those are equivalent, but people should not be surprised by the callback and need to adjust their learning rate: the same one should work with batch_size * n or batchsize and GradientAccum(n) to be as unsurprising as possible.

PR #195, opened 2020-03-23
Tidy up the use of rank_distrib and distrib_barrier.

philtrade:
As I port a few distributed-training patches from v1 to v2, also tidy things up a bit.

* `{load,save}_model`: Since `load_model` and `save_model` are both exported, and are used by `Learner.{load, save}`, the synchronization guards should probably be moved here.

* `load_learner`, `text.learner.load_pretrained`:  Use `distrib_barrier` to be consistent with other `torch.load()` callers.

p.s. first time checking in with the `nbdev` toolchain, please let me know if I missed anything.  e.g. should I include the generated `.py` files in the check-in as well ??  Thanks.


sgugger:
Thanks for that!

PR #196, opened 2020-03-23
PILDicom accepts bytes as input

Tato14:
As commented [here](https://forums.fast.ai/t/deployment-dicom-images/66288) deployment with ipywidgets returns `bytes`. In order to make `PILDicom.create` to manage it, an instance on `PILDicom` class is added.


sgugger:
Thanks for your PR. This is going to be erased the next time we build the library however: you need to make the change in the notebook mentioned at the start of that file (here medical.imaging).
Merging for this time and will port the change, but please take note for other PRs.

Tato14:
Sorry about that, first PR :smile: 
Next time I will do it the proper way

PR #199, opened 2020-03-25
Removed pos_label from CohenKappa 

jaeeolma:
Forgot to remove `pos_label` from CohenKappa last time. Also removed `labels` from both HammingLoss and HammingLossMulti (it's deprecated, doesn't do anything and will be removed in sklearn 0.23)


sgugger:
Thanks a lot!

PR #201, opened 2020-03-25
Fixes get_x and get_y for multiple blocks

lgvaz:
This PR is not ready to be merged, I'm instead submitting this to ask for advice.

The problem I'm facing is related to the example that is failing under `tutorial.datablock` (included). Here is the snippet:
```python
pets = DataBlock(blocks=(ImageBlock, CategoryBlock), 
                 get_items=get_image_files, 
                 splitter=RandomSplitter(),
                 get_y=[attrgetter("name"), RegexLabeller(pat = r'^(.*)_\d+.jpg$')],
                 item_tfms=Resize(128),
                 batch_tfms=aug_transforms())
```

In the proposed solution, I added some checks to make sure the number of functions passed are always equal to the number of inputs/targs we expect, for get_y we have the following:
```python
n_targs = len(self.getters) - n_inp
if len(L(self.get_y)) != n_targs: 
    raise ValueError(f'get_y contains {len(L(self.get_y))} functions, but must contain {n_targs} (one for each target)')
```

This causes an error to be thrown in the provided example:
```
ValueError: get_y contains 2 functions, but must contain 1 (one for each target)
```

I'm asking for advice on how should I handle this. I understand the use case, and I think it's very handy when we are using the standard (single input, single target) mode, and since this is by far the most common use case, I think this functionality should stay.

What we can do is remove the check if the number of passed blocks is 2 and n_inp is 1, and just act the same way as before. But while this would fix the specific issue, it would mean that the api behaves  bit differently depending on those variables.

Another thing we could do is, if we want to apply multiple functions to a specific block, we should pass a list. In this case we would have something like:
```python
get_y=[[attrgetter("name"), RegexLabeller(pat = r'^(.*)_\d+.jpg$')]],
```


sgugger:
Thanks for the PR. First I'd like to keep the default for `n_inp` which is not 1 but number of blocks - 1.
Then for the problem you discuss, there is always the possibility of putting the two functions in a Pipeline to make it appear as one which would remove the error.
We could also remove the check when the number of blocks for x/y is 1 (and if we get a list, pass that list as is done now) to avoid the user to write Pipeline in those cases.
Or we could not implement the functionality you are adding and solely rely on getters when there is more than one input/target.

@jph00 I'd love to hear your thoughts on this.

sgugger:
Ok, after discussion with Jeremy offline, we'll go for the Pipeline solution, so the right syntax for the example above would be:
```python
pets = DataBlock(blocks=(ImageBlock, CategoryBlock), 
                 get_items=get_image_files, 
                 splitter=RandomSplitter(),
                 get_y=Pipeline(attrgetter("name"), RegexLabeller(pat = r'^(.*)_\d+.jpg$')),
                 item_tfms=Resize(128),
                 batch_tfms=aug_transforms())
```
I'm not sure this is going to work out of the box, so I can take over and make the necessary changes if it fails. Can you just fix merge conflicts (I think the easiest way is going to be to click resolve conflicts, copy everything, paste it in a file on your machine and run `nbdev_fix_merge` on that file) and fix the default to `n_inp`? Thanks!

lgvaz:
Alright! I'll implement the proposed solution and I'll check if all the datablock examples still work =)

lgvaz:
Another tiny problem, the previous `n_inp` default was `None`, and it was being saved as `None` at the end of the constructor.
I changed it to be: `self.n_inp = ifnone(n_inp, len(blocks)-1)`

But this don't really make sense in a language model datablock, for example:
```python
imdb_lm = DataBlock(blocks=(TextBlock.from_df('text', is_lm=True),),
                    get_x=attrgetter('text'),
                    splitter=RandomSplitter())
```
`n_inp` would be 0 in this case (1block - 1)
I see `n_inp` is being changed at `Datasets`, should I leave `n_inp` as `None` in the `DataBlock`?

lgvaz:
If you think it's a good idea to assign `n_inp` in the datablock, a proposed solution is:
```python
self.n_inp = ifnone(n_inp, max(1, len(blocks)-1))
```

Which is very similar on what happens inside `Datasets`

lgvaz:
The only reason that I'm resisting to leave `n_inp` as `None` and just pass it to `Datasets` to handle, is that I need to calculate it for checking that the number of functions passed to `get_x` and `get_y` are correct

sgugger:
> If you think it's a good idea to assign `n_inp` in the datablock, a proposed solution is:
> 
> ```python
> self.n_inp = ifnone(n_inp, max(1, len(blocks)-1))
> ```
> 
> Which is very similar on what happens inside `Datasets`

That is exactly what I wanted, thanks!

PR #203, opened 2020-03-26
`BnFreeze` callback for fully frozen BatchNorm layers

rsomani95:
Replicate `BnFreeze` from v1 (https://docs.fast.ai/train.html#BnFreeze) with some additional documentation, examples and a test


rsomani95:
The notebook is a bit verbose with outputs of the `models_equal` function. Let me know if you'd like that removed, or export `models_equal` into the fastai library, or of any other changes.

sgugger:
Thanks for the PR. Don't forget the notebooks also become the documentation, so we can't have something that verbose. Just asserting it's false is enough for the test I believe.

rsomani95:
That sounds fair. I'll make that change and commit.

sgugger:
Thanks! I'll just add a #slow tag to your tests since they are a bit long.

PR #206, opened 2020-03-27
cleans code

lgvaz:
Same thing, a little bit easier to understand =)


sgugger:
In this instance, why not, but in general it's best to avoid using `ifnone` because it makes python evaluate the second argument (whereas in the test, it's not computed if not needed).

lgvaz:
> In this instance, why not, but in general it's best to avoid using `ifnone` because it makes python evaluate the second argument (whereas in the test, it's not computed if not needed).

That's a very important insight, I had not realised that so far. I couldn't help but wonder why `ifnone` was not being used everyone instead of the standard `a if a is not None else b`

PR #209, opened 2020-03-28
fixes #205

### LINKED ISSUE #205: ValueError: This DataLoader does not contain any batches
I am getting the below errors while running _24_vision.gan.ipynb and lesson7-wgan.ipynb notebooks.
image
image
Please help.

lgvaz:
I did not ran the `lesson7-wgan` notebook to not lose the outputs, but it should be working.

I remove the `_` from the files, since they are now working =)


lgvaz:
There are a bunch of changes in the files, but I think that's because it was a while since the notebooks were built. The only I think I changed was adding `ds_idx=0` to `show_results`

sgugger:
Thanks for the fixes!

PR #213, opened 2020-03-29
fix c_out->n_out

codescv:
c_out -> n_out in resnet module


sgugger:
Ah yes, thanks for catching this!

PR #214, opened 2020-03-29
Captum Callback

nareshr8:
Captum Callback Integration


nbs/73_callback.captum.ipynb
```diff
@@ -0,0 +1,215 @@
+{
```
sgugger: <p>Just a few comments on this cell: we should harmonize the names, so if it's <code>enc_inp</code>,  <code>enc_pred</code>,  it should be <code>dec_inp</code> (instead of <code>decoded_img</code>)   and <code>dec_pred</code> (instead of <code>decoded_pred</code>).</p><p>Then <code>dl.decode((self.enc_inp,self.enc_preds))</code>  should be computed only once and stored for efficiency.</p><p>Lastly, the test for colors can fit on one line:</p><p><code>self.colors = [(0, '#ffffff'),(0.25, '#000000'),(1, '#000000')] if colors is None else colors</code>.</p><p> <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/fastai2/pull/214/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

sgugger:
Thanks for adding this! I have a few cosmetic comments on the code. Let me know if you can't address them and I'll merge then polish in a follow-up commit.

nareshr8:
Thanks for the comments. I have made the changes as suggested and commited. Kindly have a look.

sgugger:
Thanks!

PR #215, opened 2020-03-29
keeps split_idx when calling _new in TfmdLists

lgvaz:
`split_idx` was not being kept when calling `_new`.  

The problem I faced because of this was: I exported my learner. Afterwards, I loaded my learner by using `load_learner`, I then created a new `DataLoader` by doing `learn.dls.valid.new`, I used valid here because I wanted only validation transforms to be applied to my data, but for some reason, `RandTransforms` were still being applied. It ended up that `split_idx` was not being saved.

If this is unclear, I can provide a working example.


sgugger:
Could you show me an example of the failing behavior? I'm not 100% sure this is the right fix. Also please use 4-space indents.

lgvaz:
Ow the indents again, I'm so sorry.. Jupyter does not gives any visual clue that I'm mixing idents so I end up forgetting. I'm going to fix that and provide an example

lgvaz:
I uploaded a minimal example to [this](https://gist.github.com/lgvaz/28f0c5a7f76a7bd1f4b210e7d0f91055) gist

sgugger:
Thanks for taking the time to explain it to me patiently, looks good to me!

PR #217, opened 2020-03-30
feat(wandb): improve callback

borisdayma:
This PR adds the following features to `WandbCallback`:

1. the callback can be used on a run with previously logged data, for example if it was instantiated multiple times on the same `wandb.init()`
1. the "raw loss" is logged
1. inputs, predictions and ground truths are logged separately -> this gives more flexibility in visualizations such as seeing inputs/predictions/labels or evolution of predictions over time
 image
1. log of config parameters -> let us see summary table of runs, filter, sort, create reports, see feature importance, etc


borisdayma:
I'm logging both loss (as "raw_loss") and smooth_loss (as "train_loss") as it can potentially help debugging. Let me know if that's ok.

Also, logging of config parameters helped me make [this report](https://app.wandb.ai/borisd13/l2ab/reports/Gradient-Accumulation-Test--Vmlldzo3MDM2MA).
I would typically log config parameters manually but it is always prone to mistakes. Having a good log of all parameters let us create automatically a [parameter importance table](https://app.wandb.ai/borisd13/lightning-kitti/sweeps/lrqc8fu3?workspace=user-borisd13) that is based on a random forest on metrics.

Right now, this callback will log:
* model input dimensions (let us check image size, etc…)
* batch size,
* model name and parameters, frozen idx…
* optimizer name, loss function name
* all the callbacks used
* all the transforms used, etc

I would like to go further and log:
* init parameters of every callback
* init parameters of learner
* type of loop (fit_one_cycle, fit…)

The easiest implementation would be to add an attribute `params_init` to Learner, callbacks, loss functions, optimizers, dataloaders (if possible). We just loop through them and if they have `params_init` we log them as `callback_name.param`, for example we will have `GradientAccumulation.n_acc`, `Learner.lr`, `Optimizer.lr` (which can help with debug…).

The alternative would be for `WandbCallback` itself to look for the right parameters based on the callback name but they are not always saved and it would be hard to maintain.

What do you think? I can try to propose something in a separate PR.

fastai2/callback/wandb.py
```diff
@@ -72,31 +83,66 @@ def after_epoch(self):
...
+        _try_set(config, 'dataset.tfms', f'{self.dls.dataset.tfms}')
+        _try_set(config, 'dls.after_item', f'{self.dls.after_item}')
+        _try_set(config, 'dls.before_batch', f'{self.dls.before_batch}')
+        _try_set(config, 'dls.after_batch', f'{self.dls.after_batch}')
+
+        wandb.config.update(config, allow_val_change=True) # in case callback runs multiple times
+
+
+# Cell
+def make_plt(img):
```
sgugger: This function could be private, I think.

sgugger:
Thanks for the PR!
For `params_init`, why not just look at the init signatures and then log the attr if it has been stored? I would rather not add a new thing to every fastai class.

fastai2/callback/wandb.py
borisdayma: Ok, if you want I can also put `_make_plt` and `_try_set` as static methods within `WandbCallback` (I don't think they would be needed for any other function).

borisdayma:
Ok, I understand completely why you would not want to add extra parameters in classes.

My main objective is for people not to have to use `wandb.config` and have it just log automatically almost everything that is possible. When you have a few runs, there's a feature that lets you clean the summary table to show only what is different.

I really like the idea of using the signatures but I see a few limitations (may be due to my limited experience with it):

* I cannot get init args from `learn` but from `Learner`. If I used `cnn_learner` which has extra init args I would not know. I could just try to get all args from all possible learner functions but it will be hard to maintain when new ones are created.
* some args are not retrievable, for example `y_range` in `cnn_learner` would be very hard, `MixUp.alpha`
* some args are set but should not be logged directly -> `FetchPredsCallback.cbs`, 
* I'd like to know if the user used `fit`, `fit_one_cycle`, `fit_flat_cos`… I think I can see if it's `fit` by not having `ParamScheduler` but it's hard to differentiate between the others.

One important parameter that people usually tweak is `lr` (not easy to find from the `lr` plot) but I think we can retrieve it from the optimizer at `begin_fit`. However, all those "hacks" may break easily as the library is updated.

Those were the main reasons I had thought of adding an optional `params_init` arg which would simplify the process but it's true that the main use would be for logging callbacks. I think it could be pretty useful though for anyone to debug.

Let's just close your other comment and I could do a separate experimental PR to test the extent of proposed changes and see the value of it (and we could just cancel that PR with no problem if it brings too much added weight). What do you think?

sgugger:
We can try, something that would give a minimal number of new lines of codes, hopefully. Maybe a decorator for functions like `cnn_learner` and such? You are welcome to experiment and suggest a PR on which we can iterate.

PR #223, opened 2020-03-30
adds unique option to show_batch

lgvaz:
Adds the option of showing the same item on `show_batch`.

I'm using this functionality so much that I thought it was a good idea to make a PR =). With this, even some examples on the book can be simplified.

`learner.py` is just an intruder in the PR xD


fastai2/data/core.py
```diff
@@ -86,10 +86,13 @@ def _pre_show_batch(self, b, max_n=9):
         if not is_listy(b): b,its = [b],L((o,) for o in its)
         return detuplify(b[:self.n_inp]),detuplify(b[self.n_inp:]),its
 
-    def show_batch(self, b=None, max_n=9, ctxs=None, show=True, **kwargs):
+    def show_batch(self, b=None, max_n=9, ctxs=None, show=True, unique=False, **kwargs):
+        old_get_idxs = self.get_idxs
+        self.get_idxs = (lambda: Inf.zeros) if unique else old_get_idxs
```
sgugger: Not need to make change if unique is false so I'd just do 
```
if unique: 
    old_get_idxs = self.get_idxs
    self.get_idxs = (lambda: Inf.zeros)
```
And test at the last line again.

lgvaz: This would throw an error at line 95 though.

Can we instead do:
```python
old_get_idxs = self.get_idxs
if unique: self.get_idxs = lambda: Inf.zeros
```

Line 95 is required right? We need to make sure to revert any changes made to `self.get_idxs`

sgugger: (Sorry looks like GitHub held my comment since last night because it was requiring an extra click)

lgvaz: Not a problem, I'll make the modifications =)

lgvaz:
Looks good? I remembered to use 4 spaces for indentation this time 😅

sgugger:
Thanks!

PR #225, opened 2020-03-31
feat(wandb): improve logging of config parameters

borisdayma:
This PR is currently marked as a work in progress.

The idea is to be able to record and log all interesting function parameters.

Currently a `@log_args` decorator is proposed.
A test is shown in notebook.

A few points:

* I think we would want to avoid some parameters so for example `Datasets.__init__ = log_args(Datasets.__init__,but='self,items,splits,tfms,tls,verbose')`
* if we use it on both `cnn_learner` (through `to_return=True`) and `Learner.__init__`, we will potentially have unexpected behavior. Maybe `cnn_learner` has a parameter set to None that is set to a value when calling `Learner.__init__`. It's probably ok though.
* I'm thinking of logging the parameters in W&B as "class.function.parameter" because `lr` can be set in so many places for example.
* I'm thinking of logging which functions are called. For example `Learner.fit_one_cycle` = True but if I add the decorator on `fit` it will also be called. Either it's ok or I can add an arg in decorator to decide when to do it.


sgugger:
What I was thinking (and it might be crazy) would be that the decorator attaches or updates an `init_arg` dictionary to the object (that's for inits or functions that return something) or attaches or updates that `init_arg` to the first argument otherwise (for fit/fit_one_cycle...).
Each time the dictionary would be function name to parameters we save (and the but argument is great to remove things we don't want), then we would just need a convenience method somewhere to that the Learner goes read the `init_arg` of its callbacks, before sending the whole thing to the logging process.

borisdayma:
> What I was thinking (and it might be crazy) would be that the decorator attaches or updates an `init_arg` dictionary to the object (that's for inits or functions that return something) or attaches or updates that `init_arg` to the first argument otherwise (for fit/fit_one_cycle...).

I think we thought of the same thing!

```python
# simple test
class tst:
    @log_args
    def __init__(self, a, b, c=3, d=4):
        pass
tst_item = tst(1,2)
assert tst_item.params_init == dict(a=1,b=2,c=3,d=4)
```

If I attach it to `fit_one_cycle`, it should update the dictionary and overwrite existing keys.
I'd like to add something to say which function was called though (when it's not `__init__`).

I've not tested it yet but it should also work with `cnn_learner` by attaching this dictionary to the returned result.

> Each time the dictionary would be function name to parameters we save (and the but argument is great to remove things we don't want), then we would just need a convenience method somewhere to that the Learner goes read the `init_arg` of its callbacks, before sending the whole thing to the logging process.

I don't understand the "function name to parameters we save". Do you mean for example `{'Learner.fit_one_cycle.lr`:0.01}`. If so, I definitely to update the way my parameters are logged so it does that as well (except for `__init__`).

I was going to add a method in `WandbLearner` that checks if there's a dict in every callback and learner (and maybe optimizer) and just merge them together. A good place may be to put it in the Learner as it could be used by other callbacks.

sgugger:
> I don't understand the "function name to parameters we save". Do you mean for example `{'Learner.fit_one_cycle.lr':0.01}`. If so, I definitely to update the way my parameters are logged so it does that as well (except for `__init__`).

Yes, I was thinking of this so that we can keep track of how parameters are replaced, for instance:
```
{'Learner': {'lr': 3e-3, 'wd': 0.01, 'moms': (0.95, 0.85, 0.95), dls': ... },
 'Learner.fit_one_cycle': {'lr': 0.01, 'wd': 0.01, 'moms': (0.95, 0.85, 0.95)}}
```
but now that I can think of it, updating the parameters with the last function called will give you the proper value.

The problem is if different callbacks have conflicting attribute names then.

borisdayma:
I think we will be able to debug by seeing the actual value plotted. It may not be perfect but this should help.

In the current version we have I already tried to get manually as many parameters as possible (that I thought were very important) and I think I have "lr" from the optimizer (if not I will add it).

So we will have "Learner.lr", "Lerner.fit.lr"... and also just "lr" coming from the optimizer.

borisdayma:
So I did a quick test by changing those 4 functions:

```python
Datasets.__init__ = log_args(Datasets.__init__, but='items,splits,tfms,tls,verbose')
Learner.__init__ = log_args(Learner.__init__, but='dls,model')
Learner.fit = log_args(Learner.fit)
Learner.fit_one_cycle = log_args(Learner.fit_one_cycle)
cnn_learner = log_args(cnn_learner, to_return=True, but='dls')
```

Then I try it on this code:

```python
path = untar_data(URLs.MNIST_TINY)
items = get_image_files(path)
tds = Datasets(items, [PILImageBW.create, [parent_label, Categorize()]], splits=GrandparentSplitter()(items))
dls = tds.dataloaders(after_item=[ToTensor(), IntToFloatTensor()])
os.environ['WANDB_MODE'] = 'dryrun' # run offline
wandb.init(anonymous='allow')
learn = cnn_learner(dls, resnet18, loss_func=CrossEntropyLossFlat(), cbs=WandbCallback())
learn.fit_one_cycle(3)
```

And here is what I get:

-> tds.init_args

```python
{'Datasets.__init__.n_inp': None,
 'Datasets.__init__.dl_type': None,
 'Datasets.__init__.use_list': None,
 'Datasets.__init__.do_setup': True,
 'Datasets.__init__.split_idx': None,
 'Datasets.__init__.train_setup': True,
 'Datasets.__init__.types': None}
```

-> learn.init_args

```python
{'Learner.__init__.loss_func': FlattenedLoss of CrossEntropyLoss(),
 'Learner.__init__.opt_func': <function fastai2.optimizer.Adam(params, lr, mom=0.9, sqr_mom=0.99, eps=1e-05, wd=0.01, decouple_wd=True)>,
 'Learner.__init__.lr': 0.001,
 'Learner.__init__.splitter': <function fastai2.vision.learner._resnet_split(m)>,
 'Learner.__init__.cbs': WandbCallback,
 'Learner.__init__.metrics': None,
 'Learner.__init__.path': None,
 'Learner.__init__.model_dir': 'models',
 'Learner.__init__.wd': None,
 'Learner.__init__.wd_bn_bias': False,
 'Learner.__init__.train_bn': True,
 'Learner.__init__.moms': (0.95, 0.85, 0.95),
 'cnn_learner.arch': <function torchvision.models.resnet.resnet18(pretrained=False, progress=True, **kwargs)>,
 'cnn_learner.loss_func': FlattenedLoss of CrossEntropyLoss(),
 'cnn_learner.pretrained': True,
 'cnn_learner.cut': None,
 'cnn_learner.splitter': None,
 'cnn_learner.y_range': None,
 'cnn_learner.config': None,
 'cnn_learner.n_out': None,
 'cnn_learner.normalize': True,
 'cnn_learner.opt_func': <function fastai2.optimizer.Adam(params, lr, mom=0.9, sqr_mom=0.99, eps=1e-05, wd=0.01, decouple_wd=True)>,
 'cnn_learner.lr': 0.001,
 'cnn_learner.cbs': WandbCallback,
 'cnn_learner.metrics': None,
 'cnn_learner.path': None,
 'cnn_learner.model_dir': 'models',
 'cnn_learner.wd': None,
 'cnn_learner.wd_bn_bias': False,
 'cnn_learner.train_bn': True,
 'cnn_learner.moms': (0.95, 0.85, 0.95),
 'Learner.fit.n_epoch': 3,
 'Learner.fit.lr': None,
 'Learner.fit.wd': None,
 'Learner.fit.cbs': (#1) [ParamScheduler],
 'Learner.fit.reset_opt': False,
 'Learner.fit_one_cycle.n_epoch': 3,
 'Learner.fit_one_cycle.lr_max': None,
 'Learner.fit_one_cycle.div': 25.0,
 'Learner.fit_one_cycle.div_final': 100000.0,
 'Learner.fit_one_cycle.pct_start': 0.25,
 'Learner.fit_one_cycle.wd': None,
 'Learner.fit_one_cycle.moms': None,
 'Learner.fit_one_cycle.cbs': None,
 'Learner.fit_one_cycle.reset_opt': False}
```

What do you think?

sgugger:
This looks like what you need to log ;) As expected there are duplicates, but we can put the burden of reading this on the user. It's better to get all the information in one place.

The next steps should be:
1. to put the log_args decorator in fastcore.utils
2. once merged make a PR that adds it to the useful stuff
3. revisit this PR to send the info to wandb
What do you think?

borisdayma:
Do you think I should make a function that collect all those parameters as part of the Learner class?

sgugger:
Yes, in step 2. Patching it at the end of the Learner notebook is probably the best place

borisdayma:
I'm just closing this PR and opening a new one to avoid unnecessary large commit history (since changes are in fastcore at the moment)

PR #229, opened 2020-03-31
fixes #228 and adds comment at end of notebook

### LINKED ISSUE #228: 10_tutorial.pets return type annotation
In the notebook `10_tutorial.pets` just after we create `SiamesePair` we have the following text:

_... This kind of filtering is done with type annotations. We indicated with the return type annotation of `SiamesePair.encodes`..._

But we are not actually using return type annotations, we are just returning the correct type. A return type annotation would be something like:
```python
def encodes(self, i)->SiameseImage:
```
I recall that fastai used to use this return type annotations, is that still happening?

lgvaz:
I basically copied some sections from fastbook chapter 11, I hope that is not a problem 😅


sgugger:
Thanks!
Though I will rework all of that in the process of rewriting it. You should rather check done tutorials ;)

lgvaz:
Alright! So I should focus on tutorials that have a green checkmark [here](https://forums.fast.ai/t/documentation-help-needed/66289)?

sgugger:
Yup

PR #232, opened 2020-04-01
replace nn.Module by fastai.Module

tcapelle:
A quick typo replacement


sgugger:
I feel this is the third time I'm merging this PR. I wonder why they keep coming back as Module.
In any case, thanks!

PR #233, opened 2020-04-01
Fix at learn.destroy()

rraminen:
Commented "learn.destroy()" in lesson3-camvid.ipynb as "del learn" already exists and Learner does not have attribute called 'destroy'
Please refer the below screenshots for reference:
image

image


sgugger:
This has been fixed yesterday already, but thanks for the PR ;)

PR #235, opened 2020-04-02
c_out --> n_out

philtrade:
ooops....


philtrade:
I also saw these:
```
./nbs/06_data.block.ipynb:    "def RegressionBlock(c_out=None):\n",
./nbs/06_data.block.ipynb:    "    return TransformBlock(type_tfms=RegressionSetup(c_out))"
```
Do the `c_out` here also need to be changed?

PR #238, opened 2020-04-03
An alternative way to do distributed training using context manager

philtrade:
Follow up to discussion in #235.


philtrade:
Hello @sgugger,

I am taking the opportunity of porting distributed training in notebook usage to `fastai v2` to see if the distributed training semantics can be cleaner, and less error prone.  In summary: I propose using a context manager to localizes/auto-clean up the scope to attach a DistributedTrainer to a Learner object.

Currently it requires "setup_distrib()", a check on `num_distrib() <1`, then call `to_distributed()` before the `fit()`.  As `DistributedTrainer` dangles around afterwards, it causes interference with other moving parts.
 
- E.g 1. in fastai v1, I spent quite some time to make LRFinder to work in distributed mode, without significant benefit in the end (only slower in fact, because can't early stop to avoid deadlock, if you recall).
- E.g. 2. in 01_intro notebook of fastai v2, after the `fit()`, the `DistributedTrainer` would interfere with the subsequent `show_results()`

Thus, to make the above work smoother, less arm-twisting, I thought `DistributedTrainer` should get out of the way when not needed.  Such "Move in, use, move out" semantic naturally lends itself to  a context manager.  So I created one `distrib_ctx(self:Learner, cuda_id)`.

I tested it with train_imagenette.py, in single and multiple GPU mode, with and without call to `setup_distrib()`, and with/without passing the `gpu` argument to the context manager (it figures it out from `rank_distrib()`.   The usage in `train_imagenette.py` is simplified, and yet is backward compatible.

Two other suggestions to run by you:

- if this context manager approach is acceptable, perhaps I can do the same to `to_parallel()` ?

- perhaps consider  a `Learner.remove_cb(class_type)`, instead of instance of the type?  Right now I have to be mindful of the snake_case version of the class name, and check `hasattr` on it first, then `remove_cb(self.snake_case)`.  That can be a separate discussion, of course...

sgugger:
Would the context manager then work in a notebook? How does that work to launch the different processes from there?

The feature is interesting in itself so we can definitely merge this. Just adapt the doc comment of `distrib_ctx` to the rest of the style of fastai v2: a one-liner in double quotes, then any additional info is in Markdown after the cell that defined the function.

We can definitely do the same thing for `to_parallel`. I'll work on the problem with `Learner.remove_cb`. I've created issue #239 to track progress on this, so when you see it's closed, you know you can simplify your code ;)

philtrade:
I'll respond the simple bits first.

> adapt the doc comment of `distrib_ctx` to the rest of the style of fastai v2: a one-liner in double quotes, then any additional info is in Markdown after the cell that defined the function.

Will do, thank you for the guideline!

> the problem with `Learner.remove_cb`. I've created issue #239

Thank you.  I will add one related idea to that issue's discussion then.

> Would the context manager then work in a notebook? How does that work to launch the different processes from there?

Yes and no.

Yes, the context manager works in notebook with single GPU as-is, because the trio `num_distrib(), rank_distrib, setup_distrib()` gracefully degenerate to base case, non-distributed mode when world size <= 1 and/or `os.environ["RANK"]` isn't set.

No, the context manager doesn't launch extra processes, but it is a stepping stone to elegantly work with other mix bag of activites: *data exploration, train, find learning rate, switch data, then train again...*.

At this moment, to launch multiple processes in notebook for v2, I'm using/porting the [ipython extension I wrote a month ago, `Ddip`](https://www.github.com/philtrade/Ddip) to v2:
```
%load_ext Ddip
%makedip -a fastai_v2 ...
... some cells later ...
 %%dip on  
... learner and data loaders created in parallel on all processes ...
with learner.distrib_ctx(): learner.fit(...)
```
 
The experience when getting `Ddip` to work with `fastai v1,v2`, had led me to believe the `with ...():` context manager approach will simplify tasks and reduce bugs.  

I haven't been able to replace `Ddip`'s functionality in a single cell context in `v2` yet.  A learning object created early on in the parent process needs to be completely "pickled"/duplicated over to children processes, and the data loaders etc, have to have correct parameters too.  Technically I haven't solved all the bugs yet, but if and when that is solved, I imagine it should *fit nicely into the context manager* semantics, with minimal code change.

philtrade:
> do the same thing for `to_parallel`. I'll work on the problem with `Learner.remove_cb`. I've created issue #239...

I see you already fixed `remove_cb()`, very fast and thank you!   I will do the context manager for `parallel`, and the `remove_cb(type)` simplification in a separate pull request, as #238 probably shouldn't depend on #239.

philtrade:
> Just adapt the doc comment of `distrib_ctx` to the rest of the style of fastai v2: a one-liner in double quotes, then any additional info is in Markdown after the cell that defined the function.

Done, per [commit 8b233d2 above](https://github.com/fastai/fastai2/pull/238/commits/8b233d2f776b7cf760ce78060151f63924a2e16a).

Let me know if need additional changes.  Thanks!

sgugger:
Thanks! I'll let you follow up with another PR to simplify the `remove_cb`.

philtrade:
> Would the context manager then work in a notebook? How does that work to launch the different processes from there?

@sgugger  Hello Sylvain, hope you're doing well and staying safe.

Regarding the above, an update I wanna share with you.  I've got this to work via a single wrapped function call, and the Jupyter session will receive the trained Learner/model.

https://github.com/philtrade/mpify/blob/master/README.md

Cheers.

PR #240, opened 2020-04-03
Link batch size in dls.from_dset

hoftherose:
Setting batchsize in "from_dset" wouldn't work. Simply needed to add batchsize to the dl_type parameters


sgugger:
Good catch, thanks!

Issue #239, opened 2020-04-03
Make `Learner.remove_cb` and friends accept types


sgugger:
Currently the process of removing a `Callback` with `Learner.remove_cb`, `Learner.remove_cbs`, `Learner.removed_cb` requires actual objects instead of types, which can yield to a bit of a gymnastic like:
```
if hasattr(learn, 'cb_camel_snake_name'): learn.remove_cb(learn.cb_camel_snake_name)
```
An easier way would be that passing types to those function also work, so you can simply call
```
learn.remove_cb(CbClassName)
```

PR #242, opened 2020-04-04
Add parallel training context manager.

philtrade:
Follow up to #238 and #239 .

distributed.py: new parallel training context manager.  Clean up remove_cb usage and added try:..finally: block.

train_imagenette.py: automatically select parallel or distributed parallel context manager.  Both can handle single GPU/single process base case.

Output from `train_imagenette.py` seeing multiple GPUs and training in data parallel mode:
```
ndim1@ndim1x99:~/Dropbox/fastai2$ python3  nbs/examples/train_imagenette.py 
epochs: 5; lr: 0.01; size: 128; sqrmom: 0.99; mom: 0.9; eps: 1e-06
Run: 0
Training in parallel_ctx context on GPU [0, 1, 2]
epoch     train_loss  valid_loss  accuracy  top_k_accuracy  time    
0         1.715690    1.811860    0.507006  0.893248        00:37      
1         1.512214    1.429912    0.634140  0.939618        00:26      
2         1.332851    1.496340    0.600000  0.944968        00:27      
3         1.265601    1.156051    0.737325  0.970191        00:27      
4         1.079385    1.008763    0.804331  0.976815        00:27      
```
and from  `-m fast2.launch` distributed training:
```
ndim1@ndim1x99:~/Dropbox/fastai2$ python3 -m 'fastai2.launch'  nbs/examples/train_imagenette.py 
Run: 0
epochs: 5; lr: 0.01; size: 128; sqrmom: 0.99; mom: 0.9; eps: 1e-06
Run: 0
Run: 0
Training in distrib_ctx context on GPU 1
Training in distrib_ctx context on GPU 2
Training in distrib_ctx context on GPU 0
epoch     train_loss  valid_loss  accuracy  top_k_accuracy  time    
0         1.873252    1.755080    0.492997  0.904507        00:21     
1         1.570249    1.663726    0.550802  0.933282        00:18     
2         1.453206    1.624517    0.584161  0.929208        00:18     
3         1.321385    1.497475    0.624650  0.938630        00:18     
4         1.180481    1.039552    0.790935  0.979628        00:18     
```


fastai2/distributed.py
```diff
@@ -114,15 +134,15 @@ def after_fit(self):
...
     return self
 
 # Cell
 @patch
 def detach_distributed(self: Learner):
     if num_distrib() <=1: return self
-    if hasattr(self, 'distributed_trainer'): self.remove_cb(self.distributed_trainer)
-    if rank_distrib() > 0 and not hasattr(self, 'progress'): self.add_cb(ProgressCallback())
+    self.remove_cb(DistributedTrainer)
+    if rank_distrib() > 0: self.add_cb(ProgressCallback())
```
sgugger, line 145:
```suggestion
    if rank_distrib() > 0 and not hasattr(self, 'progress'): self.add_cb(ProgressCallback())
```

sgugger: I think you still need the test here to avoid having the progress callback twice. (If you agree, the change is also needed in the notebook).

philtrade: Hi Sylvain, thanks for reviewing the change.

I thought about keep using the `hasattr()` check, or let `add_cb()`'s assertion to complain duplicate registration loud and clear, and reveal  the problem that "was the callback properly removed before?" 

```
def add_cb(self, cb):
        ...
        assert not old or isinstance(old, type(cb)), f"self.{cb.name} already registered"
```
What do u think?  I can put the `hasattr()` back if that's deemed more useful.

sgugger: The user won't be able to do anything about it since they are not adding the callback themselves. It's cleaner if the function does not add it if it's already there.

philtrade: Got it, thx!

philtrade: > the change is also needed in the notebook

Done.

sgugger:
Thanks!

PR #247, opened 2020-04-04
AttributeError: 'Model3' object has no attribute 'f'

rraminen:
Fix for the below error in lessson7-human-numbers.py
image

Fix: def reset(self): **self.h** = torch.zeros(bs, nh).cuda()


sgugger:
Thanks a lot for your PR!

Sadly, I can't merge this PR because of conflicts in lesson3-camvid notebook (I don't think you wanted to change anything there) and I made a mistake by using the web interface (I though delete would remove it from the PR but it wants to remove it entirely from the repo now...)

I'd suggest closing this PR, making sure you are on par with master then redo your change in lesson7-human-numbers (which we indeed want). Does that sound okay to you?

rraminen:
Sorry for the confusion. I will do as suggested.

PR #248, opened 2020-04-05
 callback

amaarora:
Add  callback


sgugger:
Thanks a lot! I will follow-up with the module build and docs (add to the sidebar etc...)
Side note: in the tests that are slow (like the `learn.fit_one_cycle` you commented out), just put a `#slow` at the beginning of the cell.

PR #249, opened 2020-04-05
feat: add log_args to common functions

borisdayma:
Add `@log_args` to common functions.

Requires merge of [fastcore PR 18](https://github.com/fastai/fastcore/pull/18).

*Notes*:

* few issues with signatures on some nbs examples, log_args outputs a message (does not fail):
  * `21_vision.learner.ipynb`, line `dls = pets.dataloaders(untar_data(URLs.PETS)/"images", item_tfms=RandomResizedCrop(300, min_scale=0.5), …)` -> got an unexpected keyword argument 'item_tfms' for `DataLoader.__init__` and `TfmdDL.__init__`. Actually `RandomResizedCrop` does not appear anywhere so I think this one is a mistake.
  * `37_text.learner.ipynb`, line `dls = TextDataLoaders.from_df(df, path=path, text_col='text', label_col='label', valid_col='is_valid')` -> got an unexpected keyword argument 'seq_len' for `DataLoader.__init__` and `TfmdDL.__init__`. I think this argument gets passed while the signature removed kwargs and there's no matching arg here. Not sure about the best option here…
* I also pushed a PR to nbdev because `@patch` needed to be directly above the function while `@log_args` needs actually to be in between. Some of the changes here are using this PR.

*Next steps*:

* create a function in Learner to gather all init_args + a few common ones (bs…) -> `Learner.gather_args`?
* cleanup repetition for log_args depending on what we decide in fastcore PR
* probably add log_args to get_language_model as well
* pull the args in `WandbCallback`
* cleanup the args we don't like


fastai2/callback/data.py
```diff
@@ -1,6 +1,6 @@
 # AUTOGENERATED! DO NOT EDIT! File to edit: nbs/14a_callback.data.ipynb (unless otherwise specified).
 
-__all__ = ['CollectDataCallback', 'WeightedDL', 'weighted_dataloaders', 'PartialDL', 'partial_dataloaders']
+__all__ = ['CollectDataCallback', 'WeightedDL', 'PartialDL']
```
sgugger, line 3: You are losing exported files. Could it be due to your change in nbdev?

fastai2/callback/data.py
```diff
@@ -12,6 +12,7 @@ def begin_fit(self): self.data = L()
     def after_batch(self): self.data.append(to_detach((self.xb,self.yb,self.pred,self.loss)))
 
 # Cell
+@log_args(but='dataset,wif,create_batch,create_batches,create_item,retain,get_idxs,sample,shuffle_fn,do_batch')
```
sgugger, line 15: Yeah, but_as would definitely be necessary...

fastai2/callback/progress.py
```diff
@@ -1,6 +1,6 @@
 # AUTOGENERATED! DO NOT EDIT! File to edit: nbs/16_callback.progress.ipynb (unless otherwise specified).
 
-__all__ = ['ProgressCallback', 'no_bar', 'ShowGraphCallback', 'CSVLogger']
```
sgugger, line 3: Another function you are losing.

fastai2/distributed.py
```diff
@@ -1,7 +1,6 @@
 # AUTOGENERATED! DO NOT EDIT! File to edit: nbs/20a_distributed.ipynb (unless otherwise specified).
 
-__all__ = ['ParallelTrainer', 'parallel_ctx', 'setup_distrib', 'teardown_distrib', 'DistributedDL',
-           'DistributedTrainer', 'distrib_ctx']
+__all__ = ['ParallelTrainer', 'setup_distrib', 'teardown_distrib', 'DistributedDL', 'DistributedTrainer']
```
sgugger, line 3: And here too

sgugger:
I think it can merged as is and improved upon. Thanks a lot for your hard work!

sgugger:
Side note, it looks like your PR removed all outputs of notebooks. Make sure you leave them as we like to keep full notebooks.

sgugger:
Oh and I found why we had those functions removed, they were patched context managers, so your fix to nbdev actually fixed that as well. So all is good :)

borisdayma:
Thanks, sounds good, I'll go on with the next steps.

Issue #246, opened 2020-04-04
RAdam ignores weight decay


black-and-blueish:
I noticed that the `wd` parameter in `Learner `does not have any effect on training when using `RAdam` optimizer. Looking at the code it seems obvious why:
```
    cbs = [weight_decay] if decouple_wd else [l2_reg]
    cbs = [partial(average_grad, dampening=True), average_sqr_grad, step_stat, radam_step]

```
Should not the last line be (same as in `Adam`):
`cbs += [partial(average_grad, dampening=True), average_sqr_grad, step_stat, radam_step]`

sgugger:
Yes, it's a mistake. Would you mind making a PR to fix this?

PR #250, opened 2020-04-05
Fix: Weight decay in RAdam

tumbleintoyourheart:
Fix [Issue #246](https://github.com/fastai/fastai2/issues/246#issue-593950729)


sgugger:
Thanks a lot!

PR #251, opened 2020-04-05
RAdam weight decay fix

black-and-blueish:


black-and-blueish:
Just noticed that this was fixed already

PR #253, opened 2020-04-05
Propagate gen_first from GANLearner to GANTrainer

sirykd:
**Issue:** `gen_first` is left unused in `GANLearner`


sgugger:
Thanks a lot!

PR #259, opened 2020-04-05
Fix weight decay in RAdam

tumbleintoyourheart:
Fix [#246](https://github.com/fastai/fastai2/issues/246). Sorry for the last failed PRs.


borisdayma:
That is strange, it looks good on my side. There was a recent update to nbdev which may not be updated on the CI side.
@sgugger should the workflow install nbdev from its git repo?

sgugger:
It was failing before this PR (the last commit is bad) so that's why.

PR #262, opened 2020-04-06
AttributeError: 'Model3' object has no attribute 'f'

rraminen:
Fix for the below error in lessson7-human-numbers.py

image

Fix: def reset(self): self.h = torch.zeros(bs, nh).cuda()


sgugger:
Perfect, thanks!

PR #263, opened 2020-04-06
fix(wandb): handle exceptions

borisdayma:
This should handle exceptions when an arg is not available.
I took advantage to fix last step not syncing as well as step number.

@sgugger maybe we can merge it to fix the issue but I believe `ignore_exceptions` would be better in fastcore.

Also I think it's useful to keep listing those args even if we will have `log_args` parameters because they serve a different purpose (init args vs parameters used during training). I think the section `_log_config` will be moved to `Learner.gather_parameters`.


sgugger:
Thanks for the PR. I will merge and move the ignore_exceptions to fastcore. I just need to think of places we use something similar and see if I need to add behavior to it.

PR #264, opened 2020-04-06
feat(log_args): use but_as arg

borisdayma:
* use but_as for `@log_args` which leads to cleaner definitions
* use `@log_args` on losses
* create `Learner.gather_args`

This PR requires [fastcore PR 20](https://github.com/fastai/fastcore/pull/20).

*Note*: _nbdev.py and captum.py don't seem in sync with the notebooks creating the build error

*Next steps:*

* integrate `WandbCallback._log_config` into `Learner.gather_args` -> I think those values are interesting even if they may be redundant because those are the ones used during training (vs init_args)
* update `WandbCallback` to log these parameters
* cleanup args we don't like


sgugger:
Thanks a lot! This is looking good :)

PR #268, opened 2020-04-06
Add Widget and Baseline options for IG

nareshr8:
Add Widgets to Visualise IG and start trying out baselines options presented in Distill


sgugger:
Your changes conflict with what is on the master branch already, you should fix the conflicts otherwise we can't merge this PR. Also I don't think the callback should return a widget, this should be something separate for users who don't have access to widgets.

nareshr8:
The widget that I built was for the Integrated Gradients. So do you recommend not having a widget or you prefer separate methods.

The code isn't actual conflict. My earlier commits are getting added to the PR every-time. I shall check that and update based on your comment.

sgugger:
I can't merge the PR because it says there are conflicts (and we can't fix those on the web interface with jupyter notebooks). I think it comes from the fact you did not run the command `nbdev_install_git_hooks` to strip your notebooks from metadata.

Please put back the cell with `# all_slow` as it's needed to put the slow flags for testing on this notebooks (since it does training of full models).

I still think the `visualize` method should exist without widgets. Widgets don't all work on collab for instance.

nareshr8:
For some reason my old PR changes are getting added to the new PR as well thus creating conflict as ReviewNB says. I will see and update accordingly.

If the widgets are accessible, we present them. Else we shall fallback to non widget thing. Would that work.

nareshr8:
Will add the `#all_slow` for sure

nareshr8:
I am removing the widgets and committing in a new PR

PR #269, opened 2020-04-06
untar_data rename issue

tcapelle:
Update untar_data to rename destination folder if it is not equal to download archive


tcapelle:
I don't get any diff...

sgugger:
Are you sure you have the latest nbdev? From all the diff I see, I think you need to update it then rebuild the library.

fastai2/data/external.py
```diff
@@ -243,4 +249,6 @@ def untar_data(url, fname=None, dest=None, c_key='data', force_download=False, e
         if _get_check(url) and _check_file(fname) != _get_check(url):
             print(f"File downloaded is broken. Remove {fname} and try again.")
         extract_func(fname, dest.parent)
+        print(f'fname: {fname}, dest: {dest}')
```
sgugger, line 252: We try to avoid being verbose. Maybe just make a test to see if a rename is necessary and just print something then?

sgugger:
Thanks for fixing the build. I just have one comment on the design, if you can address it.

fastai2/data/external.py
tcapelle: Oh I forgot to remove this...

tcapelle: I refactored the code, to check outside, I added a test. It is complicated to add an integral test without a missnamed downloadable dataset.
It is so painful that the unzip and untar methods don't give back the extrated folder.

nbs/04_data.external.ipynb
```diff
@@ -228,10 +228,10 @@
     {
```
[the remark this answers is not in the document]
tcapelle: ohh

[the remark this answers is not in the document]
tcapelle: done

sgugger:
Thanks a lot!

PR #271, opened 2020-04-06
NameError: name 'dblock' is not defined 

rraminen:
The below error happens in lesson7-superres.ipynb
image
Hence defined dblock before accessing it as a fix.

### DIFF
--- a/nbs/course/lesson7-superres.ipynb
+++ b/nbs/course/lesson7-superres.ipynb
@@ -1206,6 +1206,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "dblock = DataBlock(blocks=(ImageBlock, ImageBlock),\n",
+    "                   get_items=get_image_files,\n",
+    "                   get_y=lambda x: path_hr/x.name,\n",
+    "                   splitter=RandomSplitter(),\n",
+    "                   item_tfms=Resize(size),\n",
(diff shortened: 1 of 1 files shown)


nbs/course/lesson7-superres.ipynb
```diff
@@ -1206,6 +1206,12 @@
...
    "outputs": [],
    "source": [
+    "dblock = DataBlock(blocks=(ImageBlock, ImageBlock),\n",
+    "                   get_items=get_image_files,\n",
+    "                   get_y=lambda x: path_hr/x.name,\n",
+    "                   splitter=RandomSplitter(),\n",
+    "                   item_tfms=Resize(size),\n",
+    "                   batch_tfms=[*aug_transforms(max_zoom=2.), Normalize.from_stats(*imagenet_stats)])\n",
     "dbunch_mr = dblock.dataloaders(path_mr, bs=1, val_bs=1, path=path,\n",
     "                             batch_tfms=[*aug_transforms(), Normalize.from_stats(*imagenet_stats)])         \n",
```
sgugger, line 1216: You need to remove the `batch_tfms` from there, they should only be in the dblock.

sgugger:
Thanks! This notebook has not been update in a while, so I think there are some other problems in it. Left a comment, if you can address.

nbs/course/lesson7-superres.ipynb
rraminen: Thank you for pointing it out. I raised a new PR: https://github.com/fastai/fastai2/pull/273

PR #273, opened 2020-04-06
NameError: name 'dblock' is not defined

rraminen:
The below error happens in lesson7-superres.ipynb

image

Hence defined dblock before accessing it as a fix.


sgugger:
This is the same as #271 and does not fix the additional problem.

PR #275, opened 2020-04-07
learn.predict with list of texts

lgvaz:
Adds the possibility to pass a list of texts to `.predict`.


lgvaz:
For some reason doing:
```python
idxs.new(samples)
```
is breaking the kernel, so I instead did the following:
```python
retain_type(torch.cat([idxs_all, samples], dim=-1), idxs)
```

lgvaz:
It's now also possible to pass 2d tensors to predict =D

sgugger:
Thanks for your PR. 
I have mixed feelings on this one: `Learner.predict` does not work on a list of items across applications but mostly, I am not sure the results will be great: the padding token you pass along as never been seen during the training of the language model, so I'm not sure it gives better results (compared to separate predict calls).

lgvaz:
Excellent point about the padding tokens, the model only really sees them when training on the downstream tasks

I'm totally ok if you decide to not merge this one =)

lgvaz:
Can I ask for a bit of advice? In my application I need to perform some form of batch prediction (because of speed)

Since using padding tokens can confused the model, do you have any suggestions on how to properly do this?

lgvaz:
Actually, in my case the sequences will always have the same length 😁

sgugger:
Then in that case it's easy ;)
For a more general approach, if you had texts of different lengths and wanted to do batch predictions, my strategy would be to add some random padding between documents in the LMDataLoader so that the model has been trained on that token.

Note that it's not even used in the downstream task now, as we completely ignore padding in v2. So the model never learns anything about that token.

Re- the PR. I'm going to close this if you don't mind as this a bit too ad hoc. You can patch it to Learner for your application and gist the code + share it on the forum if you think it would be useful for other users.

PR #277, opened 2020-04-07
feat(wandb): improve log of functions

borisdayma:
Name of functions is now much nicer.

image

*Legend:*

* top -> new version
* bottom -> previous version


sgugger:
Thanks!

PR #280, opened 2020-04-07
Add Baseline Zeros and Uniform base image option to IG

nareshr8:
Add baseline image options for IG


sgugger:
Will fix the markers for test: if you don't put `#all_slow` on one cell, then the notebook should be executable with just the cells not marked as slow (which it can't here, since the learner is defined in the slow cells but used in other cells).

Thanks for the PR

sgugger:
I have cleaned up a bit (we only need to load the data once for instance). Btw, those are not callbacks since they don't change the training loop at all. They are more like interpretation objects build from a Learner after training, so I'm not sure there is a use to define as callbacks.

This way we could train the learner once and use it for every captum visualization. Will refactor this when I have a bit of time.

nareshr8:
Yes. I was also thinking about the same today. If this should move to interpretation module probably. But there is one base image geeration method (as per the distill mentioned in the notebook) which is created by averaging out the entire training dataset images. I haven't implemented that yet. But callbacks might be useful for that.

sgugger:
The general captum classes you implemented are not Interpretation objects either (this uses all predictions and losses). I think we should just refactor the common functionality in this notebook in a CaptumInterpretation object that is created from a learner.

nareshr8:
Okay. I'll try that tomorrow my time in the same notebook, if you haven't.

Issue #255, opened 2020-04-05
Callbacks aren't removed when fit a model if there is an exception


hal-314:
Hi,

I found that when I did `learn.fit(n_epchs, lr, cbs=MyBuggyCallback())`, MyBuggyCallback wasn't removed from the learner if an exception occurred. I tracked down the issue to here:
```
# nbs/13a_learner.ipynb 

    @contextmanager
    def added_cbs(self, cbs):
        self.add_cbs(cbs)
        yield
        self.remove_cbs(cbs)
```
It should be (at least as contextmanager docs says):
```
    @contextmanager
    def added_cbs(self, cbs):
        self.add_cbs(cbs)
        try: yield
        finally: self.remove_cbs(cbs)
```

I think that it's a bug. If so, I will open a PR fixing it and I'll search if there are similar bugs with the use of contextmanager

borisdayma:
I can confirm it happens to me as well sometimes.

I would definitely propose a PR if you can fix it!

aim123:
While trying to import vision package:
from fastai2.vision.all import *
the error is NameError: name 'log_args' is not defined

sgugger:
This should be fixed this way indeed, and same for `removed_cbs`, if you want to make a PR.

sgugger:
Should be fixed now. Thanks for flagging!

PR #283, opened 2020-04-07
Fix similar issues to #255

hal-314:


sgugger:
Oh thanks a lot for fixing those!

PR #284, opened 2020-04-07
Resize functions and Regression show_batch fix

tcapelle:
This pull request adds resizing functions to `vision.utils` from V1
- `resize_image`: resize one image
- `resize_images`: resize images in parallel.

Also adds decodes method to `RegressionSetup` to have `show_batch` when using `RegressionBlock`.


fastai2/vision/utils.py
```diff
@@ -44,4 +44,35 @@ def verify_image(fn):
 # Cell
 def verify_images(fns):
     "Find images in `fns` that can't be opened"
-    return L(fns[i] for i,o in enumerate(parallel(verify_image, fns)) if not o)
\ No newline at end of file
+    return L(fns[i] for i,o in enumerate(parallel(verify_image, fns)) if not o)
+
+# Cell
+def resize_image(file, max_size:None, dest=None, n_channels=3, ext=None,
```
sgugger, line 50: This should be =None. Also dest should be a require argument since this function does nothing if dest is None.

fastai2/vision/utils.py
```diff
@@ -44,4 +44,35 @@ def verify_image(fn):
...
+                 img_format=None, resume=False, **kwargs ):
+    verify_image(file)
+    img = Image.open(file)
+    imgarr = np.array(img)
+    img_channels = 1 if len(imgarr.shape) == 2 else imgarr.shape[2]
+    if (max_size is not None and (img.height > max_size or img.width > max_size)) or img_channels != n_channels:
+        assert isinstance(dest, Path), "You should provide `dest` Path to save resized image"
```
sgugger, line 57: One dest is a required argument, we can remove this line, and add a dest=Path(dest) at the beginning to be sure it's a Path.

fastai2/vision/utils.py
```diff
@@ -44,4 +44,35 @@ def verify_image(fn):
...
+            img = img.resize(new_sz, resample=3)
+        if n_channels == 3: img = img.convert("RGB")
+        img.save(dest_fname, img_format, **kwargs)
+
+# Cell
+def resize_images(path, max_workers=4, max_size=None, recurse=False,
+                  dest=Path('.'), n_channels=3, ext=None, img_format=None,
+                  resume=None, **kwargs):
+    path = Path(path)
+    if resume is None and dest == '.': resume=False
```
sgugger, line 72: More like `resume= dest != Path('.')`? Otherwise there is no default behavior.

fastai2/vision/utils.py
```diff
@@ -44,4 +44,35 @@ def verify_image(fn):
...
+    dest = path/Path(dest)
+    os.makedirs(dest, exist_ok=True)
+    files = get_image_files(path, recurse=recurse)
+    func = partial(resize_image, max_size=max_size, dest=dest, n_channels=n_channels, ext=ext,
```
sgugger, line 76: You don't need to define a partial, pass directly all the kwargs to parallel.

fastai2/vision/utils.py
```diff
@@ -44,4 +44,35 @@ def verify_image(fn):
...
+        if ext is not None: dest_fname=dest_fname.with_suffix(ext)
+        if resume and os.path.isfile(dest_fname): return
+        if max_size is not None:
+            new_sz = resize_to(img, max_size)
+            img = img.resize(new_sz, resample=3)
+        if n_channels == 3: img = img.convert("RGB")
+        img.save(dest_fname, img_format, **kwargs)
+
+# Cell
+def resize_images(path, max_workers=4, max_size=None, recurse=False,
```
sgugger, line 68: I think we should use `defaults.cpu` here for `max_worker`.

fastai2/vision/utils.py
tcapelle: why you don't use typing anymore?

fastai2/vision/utils.py
tcapelle: I don't really get what is resume here, the idea is that a file is already there, it should not be copied, hein?

fastai2/vision/utils.py
```diff
@@ -44,4 +44,37 @@ def verify_image(fn):
...
+def resize_images(path, max_workers=defaults.cpus, max_size=None, recurse=False,
+                  dest=Path('.'), n_channels=3, ext=None, img_format=None,
+                  resume=None, **kwargs):
+    "Resize files on path recursevely to dest to max_size"
+    path = Path(path)
+    if resume is None and dest != Path('.'): resume=False
+    dest = path/Path(dest)
+    os.makedirs(dest, exist_ok=True)
+    files = get_image_files(path, recurse=recurse)
+    func = partial(resize_image, )
```
sgugger, line 78: This line can be removed now :)

fastai2/vision/utils.py
sgugger: Yes, since resizing can be a long process.

fastai2/vision/utils.py
sgugger: We don't, no, unless for patching/type-dispatching.

fastai2/vision/utils.py
tcapelle: ups yea

sgugger:
Oh, I missed one thing at my last review: your PR changes the images, so I guess you tested the function on them without using a different folder to store the modified images. Could you remove the modifications (checkout master, copy the images then checkout your branch and replace the images by what you copied on master).

tcapelle:
> Oh, I missed one thing at my last review: your PR changes the images, so I guess you tested the function on them without using a different folder to store the modified images. Could you remove the modifications (checkout master, copy the images then checkout your branch and replace the images by what you copied on master).

Yes my bad, I discovered the temporary folder trick later on.
Restored original files.

sgugger:
Thanks a lot!

tcapelle:
Quick qeustion, what I should do with the branch once the PR is merged, just delete it or merge it with my master before delete?

sgugger:
You can delete it and recreate a new one for future PRs.
Or there is some way to properly use git to make it up to date with master but that's out of my git competences (I use the first when making PRs on other repos.)

borisdayma:
<p>I like <a href="https://cli.github.com/" target="_blank">Github CLI</a>:</p><ul><li>clone original repo (no need to make a fork)</li><li>create a new branch and commit in it</li><li>gh pr create</li></ul><p> </p><p>Then you just go back to master, update and create a branch for each PR.</p><p> </p>

PR #290, opened 2020-04-08
i is an index to the current length into the sequence. This should be…

dereknewman:
… compared to ensure it is less than the self.max_len. Fix from current state of comparing from (sl - i) which would be the remaining length.

This took me a while to diagnose what was happening, but I think the current state is comparing the remaining length of the sequence (i.e. sl - i) to the max_len, which I'm pretty sure is not what you want.

This result in this 'if' statement never being run if max_len is set, as i is always started at 0 (i.e. i in range(0, sl, self.bptt)

This causes an error because no data is in 'outs' or 'masks'


sgugger:
No this is what we want. If you have prepared your data with `pad_input_chunks`, the big majority of padding is applied first, and you want the model to have read all your texts. This test is to apply backpropagation to the *last* `max_seq_len` tokens (not the firsts), which is why we use `sl - i`: since the first outputs are not used later on in the model, it does not generate gradients. But thanks to the hidden state, the end of the output is still the same, whatever `max_seq_len` might be.

It's common to truncate backpropagation this way, but I have never heard of reading the full texts then applying backpropagation to the first tokens even if you read until the end.

PR #294, opened 2020-04-10
DataBlock.summary(show_batch=True) as a shortcut to dls.show_batch()

philtrade:
Shorten the often-repeated 3-line debugging idiom of seeing the batch from:

1. tweak `DataBlock(...... long list of options ...)`
2. create `DataLoaders`
3. then call `dls.show_batch(<some args>)`

to
1. tweak `DataBlock(....)`
2. call `DataBlock.summary(path, show_batch=True, **some_args)`

Previously I had a quick discussion with @lgvaz on how to refactor the common 3-step `dls.show_batch(...)` to something simpler.

Then I saw `DataBlock.summary()`, which is **the** suggested debugging tool for `DataBlock` in `fastbook`, and already creates a transient `DataLoaders` object --- that would be a convenient spot to add more convenience to the common debugging task of tweaking `DataBlock` and  *just want to see the effect on one batch`*, wouldn't it?


philtrade:
p.s. I also added example documentation in the nb -- wonder if it might be useful to fastbook's `05_pet_breeds.ipynb ` as well, just an idea.

sgugger:
Thanks, this sounds useful!
Note that it won't change the output in the notebook 05 of fastbook since that summary fails.

tumbleintoyourheart:
Now does the `DataBlock.summary()` create `DataLoaders` for us? If so then how do we use them? I used to do `dls = DataBlock.dataloaders()`

sgugger:
No. The just ends by printing a `show_batch`.

philtrade:
> Note that it won't change the output in the notebook 05 of fastbook since that summary fails.

Ah yes, you're right, it won't change that case.  I wonder if it maybe useful to add an example  in the 05 notebook, to demo this shortcut `summary(.... show_batch=True, ncols=..., nrows=...)` .
 
@tumbleintoyourheart : 
> Now does the DataBlock.summary() create DataLoaders for us

`summary()` was designed as a debugging aid I believe, and not for constructing a `DataLoaders` object, not to mention the multiple ways to build a `DataLoaders`.
This change is to tag along the debugging tool of `show_batch()` with that of `summary()`'s.

If you don't care about the `DataBlock()` object nor the `summary()` debugging output, and only need to use the `DataLoaders` object, and consider appending the call to `.dataloaders(...)`  to `DataBlock(.... long list of options ...)`:

```
# get handles to `DataLoaders`, then `Learner`, and disregard the `DataBlock` object
dls = DataBlock(blocks = (ImageBlock, CategoryBlock),
                 get_items=get_image_files, 
                 splitter=RandomSplitter(seed=42),
                 get_y=using_attr(RegexLabeller(r'(.+)_\d+.jpg$'), 'name'),
                 item_tfms=Resize(460),
                 batch_tfms=aug_transforms(size=224, min_scale=0.75)).dataloaders(path/"images)
learn = cnn_learner(dls, resnet34, metrics=error_rate)
```

Or, if you only need the `DataBlock` object, and don't care about the `DataLoaders` object, you may consider calling `DataBlock.dataloaders(...)` only when you are ready to use the `DataLoaders`, such as when creating the `Learner` object:

```
# get handles to `DataLoaders`, then `Learner`, and disregard the `DataBlock` object
pets = DataBlock(blocks = (ImageBlock, CategoryBlock),
                 get_items=get_image_files, 
                 splitter=RandomSplitter(seed=42),
                 get_y=using_attr(RegexLabeller(r'(.+)_\d+.jpg$'), 'name'),
                 item_tfms=Resize(460),
                 batch_tfms=aug_transforms(size=224, min_scale=0.75))
pets.summary(... do some debugging, maybe show_batch=True,.....)

learn = cnn_learner(pets.dataloaders(path=....), resnet34, metrics=error_rate)
```

sgugger:
> I wonder if it maybe useful to add an example in the 05 notebook

No, that would make the notebook be out of phase with the book now. We can have examples in the docs and the tutorials.

philtrade:
Got it, thx!

PR #296, opened 2020-04-11
Implement batch size finder

hal-314:
It's the fastai2 implementation of the "An Empirical Model of Large-
Batch Training" paper (https://arxiv.org/abs/1812.06162). See https://forums.fast.ai/t/batch-size-finder-for-fastai-v2/67790


jph00:
Sorry for the delay on this. I think this would be best done as a separate repo - perhaps you could create a new project for this?

hal-314:
For those that find this PR useful, I implemented this as a separated project [here](https://github.com/hal-314/fastai-batch-size-finder) before doing this PR. It was tested with fastai2 0.0.16.

PR #300, opened 2020-04-14
[fix] auto generated vocabulary didn't use user-defined special tokens but always default special tokens

richardyy1188:
When pass user-defined text special tokens, `make_vocabulary` still insert only default special tokens
```
db = DataBlock(
        blocks=TextBlock.from_df(
            ...
            special_toks=(defaults.text_spec_tok+['[MASK]']), 
            vocab=None,
         ),
        ...
    )
dls = db.dataloaders(df)
'[MASK]' in dls.vocab # False
```


sgugger:
Thanks for fixing this. There are a number of issues though (the main one is you should never, absolutely never, have a default argument that is a reference type like a list), will fix in a follow-up commit.

PR #307, opened 2020-04-15
Reduce LR on plateau with minimum LR

edugp:
Adds a new parameter `min_lr` to ReduceLROnPlateau to stop reducing the learning reate below its value.


sgugger:
This seems useful, thanks!

PR #310, opened 2020-04-17
feat(wandb): improve logging of config parameters

borisdayma:
Following changes have been implemented:

* log slices (example with `Learner.fit(1, lr=slice(0.05))`
image
* improve logging of config parameters (try to log all at once and if it fails, one by one)
* handle cases where we don't succeed in logging prediction samples (either when creating dataloader or when performing predictions)
* fix for bug with dataframes as [reported in forums](https://forums.fast.ai/t/fastai-v2-chat/53518/1083?u=boris)


sgugger:
All looking good, thanks!

PR #314, opened 2020-04-19
fix typos and add clarity on nbs/24_tutorial.siamese.ipynb

riven314:
among those changes, the following change aims to have a better clarity of the paragraph: 

**Before**
 "It ends up with 512 features, so for our custom head, we will need to multiply this by 4: 2 because we have two images concatenated, and 2 because of the fastai concat-pool trick (we concatenate the average pool and the max pool of the features). The `create_head` function will give us the head that is usually used in fastai's transfer learning models."
**After**
    "It ends up with 512 features, so for our custom head, we will need to multiply this by 4 **(i.e. 2*2)**: 2 because we have two images concatenated, and **another** 2 because of the fastai concat-pool trick (we concatenate the average pool and the max pool of the features). The `create_head` function will give us the head that is usually used in fastai's transfer learning models."

**Rationale**
The original sentence "we will need to multiply this by 4: 2" may lead to a misunderstanding that the colon in "4: 2" is a ratio. After the change, the colon is clearer to be an elaboration of where the factor of 4 comes from.


sgugger:
Thanks a lot!

PR #315, opened 2020-04-20
BUG: ResBlock with stride>2

scottclowe:
The pooling kernel on the identity block needs to match the stride on the conv block so the shapes match up. Previously it was hard-coded to be a pooling kernel of 2 if the conv stride was not 1.

Migrated to fastai2 from PR https://github.com/fastai/fastai/pull/2546


sgugger:
Thanks for making the fix on both repos!

PR #318, opened 2020-04-20
Reestructuring imports

WaterKnight1998:
The imports of this submodule are following now the format of the rest


WaterKnight1998:
The previous changes are incorrectly.

PR #319, opened 2020-04-21
Distrib training for tabular data

philtrade:
DistributedDL.create_item(): Only return idx here, let TabDataLoader.create_batch() do its job using iloc.
train_tabular.py: Test driver for training url.ADULT_SAMPLES from cmdline. Derived from train_imagenette.py.

torch_core.distrib_barrier(): Check against uninitialized distrib environment b4 entering synchronization barrier.


sgugger:
Thanks, could you just adapt the arguments in your examples? You are not using most of them so there is no need putting them there.

sgugger:
Thanks!

PR #325, opened 2020-04-22
Add resample param to resize methods

tcapelle:
add support to resize for other resample methods, default is Image.BILINEAR


sgugger:
Sounds good to me, thanks!

PR #328, opened 2020-04-25
TTA callback fix

bearpelican:
Fixes: `AttributeError: 'Learner' object has no attribute 'values'` when running TTA on a new learner.

Looks like event.begin_fit doesn't happen if inner=True


sgugger:
I think it should only been done once at the beginning and then have a try for all the rest and
```
except CancelFitException:             self('after_cancel_fit')
finally:                               self('after_fit')
```
at the end.

bearpelican:
Ah good point. Forgot get_preds should be included too - 40436de
Feel free to close this and push the correct fix - if that's easier.

sgugger:
Thanks!

PR #329, opened 2020-04-25
New example training script for IMDB classifier, and safe guard untar_data() in distributed training. 

philtrade:
rank0_first(f): executes `f()` on rank-0 proc then the rest.  Demo usage in new example training script `train_imdbclassifer.py`, to avoid data corruption on (re)fresh download via `untar_data()`

DistributedTrainer.fup: allow fiddling of PyTorch's  `DistributedDataParallel(...find_unused_parameters=..)`


nbs/examples/train_imdbclassifier.py
```diff
@@ -0,0 +1,50 @@
...
+
+        learn = rank0_first(lambda: text_classifier_learner(dls, AWD_LSTM, drop_mult=0.5, metrics=accuracy))
+
+        if dump: print(learn.model); exit()
+        if fp16: learn = learn.to_fp16()
+
+        # TODO: DataParallel would hit floating point error, disabled for now.
+        # if gpu is None and n_gpu: ctx = partial(learn.parallel_ctx, device_ids=list(range(n_gpu)))
+
+        if num_distrib() > 1 and torch.__version__.startswith("1.4"): DistributedTrainer.fup = True
```
sgugger, line 45: What does this line do?

sgugger: Ah ok, I understand now. The check of pytorch version should probably be done in `DistributedTrainer` then (we will probably pin fastai v2 on 1.5.0 when it's time for release so it won't be needed in the end).

philtrade: I hesitated to put the hack inside the library itself, because it only happens only under 
specific conditions:

1. it fails with `text_classifier_learner` only --- `train_imagenette.py` and  `train_tabular.py` are fine.  **And**,
2. in distributed mode only.  Single GPU, non-distributed mode is fine.  **And**,
3. it fails at the second `fit_one_cycle()` inside `fine_tune()`, after `unfreeze()`.   `fit_one_cycle(epochs, lr)` without `unfreeze()` is fine.
4. In the past: PyTorch 1.2 + distributed training + fastai v1 + IMDB classifier worked *fine*.  But as soon as I upgraded to PyTorch 1.3 or 1.4, it blew up.

Only when altogether: distributed mode + `text_classifier_learner` + `unfreeze(), fit_one_cycle()` + PyTorch 1.4, it fails.  And somehow this `DistributedDataParallel(find_unused_parameters=True)` works as advertised in PyTorch's 115-words long `RunTimeError`, but I hesitate to apply it to all learners -- "don't fix it if it ain't broken".

I haven't had time nor the expertise in PyTorch yet to debug this hairy corner case yet.  So I thought patching in the training script for now.  Like you said, PyTorch 1.5 may or may not fix it (or break other learners, touch wood! ), if so, the library needs not to be touched in the end.

Perhaps I add a comment line in the training script, explain this is a workaround for a specific bug under specific conditions? 

Thank you for the feedback!

sgugger: Ok, this is weird so until it's investigated more, it makes sense to have it here and not in the library. Thanks for your explanation!

sgugger:
Thanks a lot!

nbs/examples/train_imdbclassifier.py
philtrade: Thank you!

PR #334, opened 2020-04-29
Add x_block optional param for ImageDataLoader.from_df 

nareshr8:
Add X-Block so that different Channeled Image can be used as input with this API


nareshr8:
Added a method to pass x_block which can be used for cases like monochrome images.

Just a wild though, should we just randomly pull all images (say a given batch_size) and check if the image is monochrome (i.e) RGB channels have same values to automatically implicitly consider this as monochrome instead of having to explicitly passing the x_block..

sgugger:
No this is not the goal of the factory methods: they don't aim at being universal, just at being easy for beginners. The real API to use as soon as you know your way around the library is the data block API.

PR #336, opened 2020-04-29
df_shrink(), df_shrink_dtypes(): Utility to reduce DataFrame memory footprint.

philtrade:
A utility function to reduce DataFrame memory footprint by casting numeric columns to smallest fitting types, categorifying `object` type, and a couple other features.

Reimplementation of fastai v1's `utils/mem.reduce_mem_usage(df)` but fix/enhance a few things.
* numeric type min/max check bug fixed, should be `<=` and `>=` instead of `<` and `>'.
* add option (`int2uint=True`) to convert `int` columns to use `uint`, if column.min() >= 0.
* allow disabling the object to category casting, with `obj2cat=False`.
* allow explicit exclusion by columns names with `skip=[col1, col2...]`.
* performance improvement: on Rossmann data, `df_shrink()` runs under 700ms, down from the 2.7s the v1's `reduce_mem_usage()` version, about 75% faster.

Test cases, and usage example are provided in the `40_tabular.core.ipynb`.

Regarding this large diff of the notebook `40_tabular.core.ipynb`, I'm pulling my hair-out a bit on this.  My browser/jupyter when saving the file after executing, changes the HTML tags of table output: `<th>0</th>` became `<td>0</td>`... very annoying but I don't know how to fix these html/browser/jupyter mess.   The code that `nbdev_build_lib` generated should be good though.  Any idea what I'm doing wrong here?  Thanks...


muellerzr:
We should already do this, see `reduce_cats` and `reduce_conts`, how does it compare vs that implementation? When I was experimenting I found they are roughly the same. Also to reduce the memory even further you should do `inplace=True` to not make a copy.

philtrade:
> We should already do this somewhat, see `reduce_cats` and `reduce_conts`, How does it compare vs that implementation? 

*That implementation* does not handle integer types, nor the case where a float128 column may have values > float32 and less than float64.  Does that implementation's "self[column-names].astype()"  use column-by-column `astype()`, which is 3X slower than DataFrame.astype(dtypes) in a single shot approach?

Does  *that implementation* allow user to selectively opt-out columns?  `df_shrink()` calls `df_shrink_dtypes()` to obtain a new schema first.   Users can take a look columns can be shrunk, if they want to,  without involving any column data copying/casting.

DataFrame.astype() does not support `inplace`, but if user move to use the new copy and the old copy would be garbage collected when ref-count drops to 0, no?  Sure maybe the "peak mem usage" may go up, which may be thrash if a huge dataset must be loaded into memory all at once.

`df_shrink()` is a general routine applicable to any DataFrame, *that implementation* is specific to Tabular (or TabularPandas).  My goal/wish is to make `df_shrink()` works well with cuDF/cuPY down the road as well, thus deliberately chose an interface which only assumes DataFrame, not a class of any kind above it, like Tabular and TabularPandas.

> When I was experimenting I found they are roughly the same.

Same in terms space savings or runtime?  Which are you comparing to which?

> (`reduce_memory` is the parameter on `TabularPandas` and it's enabled by default)

I'm well aware this `reduce_memory` flag.  And precisely because I feel it's not general enough, I went on to implement this for efficiency and flexibility, squeeze out extra room (e.g. int64 -> int8/uint8), and to improve upon v1's `reduce_mem_usage`, not withstanding the fact that, depends on the dataset, most saving could be from categorifying the 'object' type.

muellerzr:
I was mostly looking at storage space, not runtime. Great work :) My recommendation would be to replace where the call to `reduce_cats` and `reduce_conts` to instead do your `df_shrink` so it's done in the `TabularPandas` object itself. 

It's not an inconsistency, at least I don't think. Sylvain was the one who advised me do that. `self.train` is initialized when you do the `super().__init(df)` so we always have it. (and just whether or not is shuffled/dropped is what happens) I believe. @sgugger can comment if my presumptions are wrong there.

philtrade:
> `self.train` is initialized when you do the `super().__init__(df)`

I see, nothing to worry there then.  Thanks for clarifying.

>  the call to `reduce_cats` and `reduce_conts` to instead do your df_shrink  so it's done in the TabularPandas object itself.

Perhaps. If and when `df_shrink()` is considered to be useful enough, the user/maintainer can decide ;)

philtrade:
Alright I realized three things:

-  `Tabular.reduce_cont and reduce_cat` do require explicit column names, hence it allows users to opt out which columns, answering my own question. 

- `DataFrame.astype()` does have a `copy=False` flag, which may just be equivalent to `inplace=True` in the `column.astype`.  Will experiment.

- re: cuDF, its current `DataFrame.astype()` implementation doesn't accept a dictionary of dtypes of "column name: style" key-value pairs.  Only a single dtype is allowed. 🤦🏻‍♂️

sgugger:
This looks very useful, thanks! From what I see and understand of the discussion, this does more than the current `reduce_cat`, `reduce_cont`? I like having at it as a function directly available to users but we could have the `reduce_memory` argument call this function instead of the current `reduce_cat`/`reduce_cont`?

This can be done in a follow-up PR, merging this as a start.

philtrade:
> This looks very useful, thanks! From what I see and understand of the discussion, this does more than the current `reduce_cat`, `reduce_cont`? 

Yes.

> we could have the `reduce_memory` argument call this function instead of the current `reduce_cat`/`reduce_cont`?

Yes, `reduce_memory` can be lazy and layback to only call `df_shrink(df)`, or be a control-freak/type-nazi by first calling `new_dtypes = df_shrink_dtypes(df)`, nick-picking on what columns to cast, then call `df.astype(nickpicked_new_dtypes)` --- exaggeration of course it is, but you get the idea.

Thank you all for da feedback.

PR #337, opened 2020-04-29
monkey patch 'siampredict' to 'Learner', also modified SiameseImage.show

yijinlee:
Hi,

I briefly explained what I did in [this](https://forums.fast.ai/t/24-tutorial-siamese-ipynb-monkey-patch-siampredict-method-to-learner/69999) post.

Modified "24_tutorial.siamese.ipynb" to make it easier to look at prediction results, by monkey-patching a `siampredict` method to the `Learner`, so that when a `SiameseImage` is passed to it, it will also show the two images and the resulting prediction of “Similar” or “Not similar”.

Also modified the `SiameseImage.show` method so that for a `SiameseImage` that was created without a “label” (i.e. without the 3rd argument of True/False or “similar”/“not similar”), when `show()` is called it will still show the two images, with title “Undetermined”.

I am not sure if this is the correct way to patch things in a notebook, and/or whether these changes are useful for the tutorial notebook or not. Happy to make further changes, if you could please suggest any. Thanks.

Yijin


sgugger:
Thanks, this looks nice!

PR #338, opened 2020-04-30
fix TensorMask ->PILBase create

tcapelle:
add TensroMask case for PILBase.create method


sgugger:
Careful, you're missing an if at the beginning of your statement ;)

tcapelle:
Je suis désolé !

sgugger:
Thanks!

tcapelle:
there was a typo

PR #344, opened 2020-05-03
Fix Tabular CategoryBlock ignoring vocab

black-and-blueish:
I noticed that when I try to pass a vocab in TabularPandas via CategoryBlock it's being ignored - instead, vocab is always recreated from y data.

This PR should fix this.


sgugger:
This looks reasonable, thanks!

PR #345, opened 2020-05-04
add loss func to RegressionSetup

tcapelle:
This PR adds loss_func to RegressionSetup, default MSELossFlat


sgugger:
Sounds reasonable, thanks!

PR #353, opened 2020-05-13
GitHub Actions: Update checkout and setup-python, use Python 3.8

EwoutH:
A bit of GitHub Actions maintenance.
 - Update to [checkout](https://github.com/actions/checkout) v2 (performs faster)
 - Update to [setup-python](https://github.com/actions/setup-python) v2
 - Use Python 3.8 (latest stable version)


sgugger:
Thanks! If you're up to it, this can also be updated on fastcore/nbdev/nbdev_template/fastprogress :)

PR #356, opened 2020-05-16
Valid dl of custom DataLoader won't get pad_input_chunk.

richardyy1188:
# What happened
I create a new `Dataloader` called [TextDataLoader](https://forums.fast.ai/t/pretrain-mlm-and-fintune-on-glue-with-fastai-2-textdataloader-for-sliding-window-over-sentences-or-tokens/71241), and `dls.train` works well, but `dls.valid.one_batch()` didn't pad the sample and will complain about something like `torch.stack can't stack tensor with different shape ...` 

# Why
Where `pad_input_chunk` passed
```
class TextBlock(TransformBlock):
...
    def __init__(self, tok_tfm, vocab=None, is_lm=False, seq_len=72, **kwargs):
        return super().__init__(...
                                dls_kwargs={} if is_lm else {'before_batch': partial(pad_input_chunk, seq_len=seq_len)})
```
And then `pad_input_chunk` is passed via `kwargs`
```
class DataBlock():
...
    def dataloaders(self, source, path='.', verbose=False, **kwargs):
        dsets = self.datasets(source)
        kwargs = {**self.dls_kwargs, **kwargs, 'verbose': verbose} # <---- pad_input_chunk is in self.dls_kwargs
        return dsets.dataloaders(path=path, after_item=self.item_tfms, after_batch=self.batch_tfms, **kwargs)
```
Here comes the problem. When creating validation dataloader, we didn't pass `kwargs` which contains `pad_input_chunk` but only `self.dl_kwargs` which is just `{}`
```
class FilteredBase:
..
    def dataloaders(self, bs=64, val_bs=None, shuffle_train=True, n=None, path='.', dl_type=None, dl_kwargs=None,
        ...
        dl = dl_type(self.subset(0), bs=bs, shuffle=shuffle_train, drop_last=drop_last, n=n, device=device,
                     **merge(kwargs, dl_kwargs[0]))
        dls = [dl] + [dl.new(self.subset(i), bs=(bs if val_bs is None else val_bs), shuffle=False, drop_last=False,
                             n=None, **dl_kwargs[i]) for i in range(1, self.n_subsets)] 
                                 # !! ￪￪ problem ￪￪ !!
        return self._dbunch_type(*dls, path=path, device=device)
```
So also pass kwargs to valid dataloader creation.

I don't know why validation `SortedDL` will pad, but after I modify it to pass also `kwargs`, my custom dataloader will work well.


sgugger:
I'll have to take a closer look on Monday but the `.new` method will use the same kwargs as the ones in the dataloader, so this should not been necessary.

sgugger:
Looking more closely, the problem comes from your `new` method in `TextDataLoader`. It does not call the superclass new, which means you lose all the state of the dataloader you apply it to. 

Look at what `LMDataLoader` does for instance, extracting the state of custom args that are not present in `DataLoader` and calling the superclass new.

PR #358, opened 2020-05-17
Let Categorize maintain order and fix `reduce_memory`

muellerzr:
Currently `Categorize` will always sort your classes, even if an order is given. This fixes that to allow for a custom order.


muellerzr:
Don't merge this yet, as it involves Tabular I'll also add in the new way to reduce

muellerzr:
Good to go now, let me know if I missed anything or if you believe this could be refactored better

fastai2/data/transforms.py
```diff
@@ -217,12 +217,13 @@ def __eq__(self,b): return all_equal(b,self)
 class Categorize(Transform):
     "Reversible transform of category string to `vocab` id"
     loss_func,order=CrossEntropyLossFlat(),1
-    def __init__(self, vocab=None, add_na=False):
+    def __init__(self, vocab=None, sort=True, add_na=False):
         self.add_na = add_na
-        self.vocab = None if vocab is None else CategoryMap(vocab, add_na=add_na)
+        self.sort = sort
```
sgugger, line 222: Since there are two of them now, please use `store_attr(self, 'add_na,sort')`.

muellerzr:
Should be adjusted now!

sgugger:
Thanks!

PR #359, opened 2020-05-17
add  (recompute_scale_factor=True)

MahmoudAlDinnawi:


sgugger:
We explicitly pass the `scale_factor`, so I don't think we want to recompute it.

PR #361, opened 2020-05-18
Expanding the Tabular documentation

muellerzr:
This is a little bit of documentation for tabular, as this has been a problem I notice people get confused on.


muellerzr:
I'll also try to flesh out some work on the tabular documentation over the next few days, if you feel this would be better suited for a different notebook let me know!

muellerzr:
@sgugger let me know if this is a good start on expanding some of the documentation? If so I'll continue :)

sgugger:
This looks good to me.
Sidenote: I think `Categorify` and `NormalizeTab` are legacy and not used anywhere, so I'll probably remove them soon.

muellerzr:
If I’m understanding correctly, instead we use the type dispatch call instead for them? (@Normalize, etc)

sgugger:
Yes, we use Categorize and Normalize.

muellerzr:
Got it, thanks!

PR #362, opened 2020-05-20
Easier to choose whether to show pad token and what is pad token need to hide

richardyy1188:
[Thread](https://forums.fast.ai/t/text-the-reasons-why-i-want-to-show-pad-in-show-batch/71688/4)

1. We can choose show pad or not easierly.
2. Even if your pad is not `xxpad`, we still can choose not to show it(e.g. `pad_token=‘[PAD]’`)

I set default `pad_tok=None`, which is not hide anything, it is good to me you create a follow up commit and change the default value if needed.


richardyy1188:
It seems I made a wrong pr, close and open a new one.

PR #368, opened 2020-05-22
Fix CutMix lerp weight bug

aaerox:
x1,y1,x2,y2,w,h are all integers, so division by W*H always results in 0.0, and so lerp amount is always 1.


sgugger:
Thanks!

PR #369, opened 2020-05-22
Fix RocAucMulti metric test

morganmcg1:
The x,y of the test needed to be at least of len 20 for compute_val to work


morganmcg1:
Closing as this fix is also rolled into https://github.com/fastai/fastai2/pull/370

PR #372, opened 2020-05-22
feat(wandb): custom logging of semantic segmentation

borisdayma:
Allow custom logging of semantic segmentation samples.

You can create custom graphs:

* select which classes to display
* add optionally input underneath the mask
* set up opacity of mask and image

 

You can even plot evolution of prediction over time.

 

See a [sample run](https://app.wandb.ai/borisd13/demo_config/runs/51wrd6ix?workspace=user-borisd13)


sgugger:
Thanks a lot, this looks very good!

PR #375, opened 2020-05-25
Fixed faulty `BnFreeze` implementation and test

rsomani95:
There was a bug in the earlier implementation where since `deepcopy(dls)` was being used for both the `Learner` objects being compared, the comparison wasn't valid. When using separate `dls`, the test failed, and I realised that `set_bn_eval()` needs to be called when the batch begins rather than when the epoch begins.


jph00:
Why does set_bn_eval() need to be called when the batch begins rather than when the epoch begins?

rsomani95:
I can't point out what exactly causes it, I suspect one of the default callbacks is setting bn layers to eval?

I've put together a [notebook](https://nbviewer.jupyter.org/gist/rsomani95/ea6b3481003ca547b1be386fcb9e496a) that displays this precisely. Didn't add this to the PR since it's a bit verbose; will be happy to change that.

jph00:
I changed the tests and it seems the old way worked OK AFAICT.

PR #377, opened 2020-05-25
ColReader breaks for colname "name"

sutt:
This is weird one: `ColReader` does not work on dataframe columns named `name`. 

This occurs because pandas also has an attribute named `.name` which confuses the fastai method of getting the column via `getattr`. 

Unfortunately, `name` is a pretty common thing to name your column, including fastai2's own MNIST_TINY labels.csv. This simple fix makes it work.

This kaggle notebook shows a detailed exploration: https://www.kaggle.com/stillsut/fastai2-repro-2

**Code details:** it looks tempting to just get rid of `getattr` here, and replace with pandas centric indexing. But other tests rely on getattr and break when we do just do `r[c]`.


sgugger:
Thanks for adding the test and good catch! Can we just have it in one line like this?
```
o = r[c] if isinstance(c, int) else r[c] if c=='name' else getattr(r, c)
```

sutt:
@sgugger I'm on it

sutt:
@sgugger All set on the correction.

sgugger:
Thanks!

Issue #156, opened 2020-03-08
SaveModelCallback cannot be used with WandbCallback


borisdayma:
`SaveModelCallback` cannot be used with `WandbCallback`.

**Reproducible example**

```python
from fastai2.vision.all import *
path = untar_data(URLs.MNIST_TINY)
items = get_image_files(path)
tds = Datasets(items, [PILImageBW.create, [parent_label, Categorize()]], splits=GrandparentSplitter()(items))
dls = tds.dataloaders(after_item=[ToTensor(), IntToFloatTensor()])

os.environ['WANDB_MODE'] = 'dryrun' # run offline
wandb.init(anonymous='allow')
learn = cnn_learner(dls, resnet18, loss_func=CrossEntropyLossFlat(), cbs=[WandbCallback(), SaveModelCallback()])
learn.fit(1)
```

**Error**

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/Projects/fastai2/fastai2/learner.py in fit(self, n_epoch, lr, wd, cbs, reset_opt)
    191                         self._do_epoch_train()
--> 192                         self._do_epoch_validate()
    193                     except CancelEpochException:   self('after_cancel_epoch')

~/Projects/fastai2/fastai2/learner.py in _do_epoch_validate(self, ds_idx, dl)
    176         finally:
--> 177             dl,*_ = change_attrs(dl, names, old, has);       self('after_validate')
    178 

~/Projects/fastai2/fastai2/learner.py in __call__(self, event_name)
    122 
--> 123     def __call__(self, event_name): L(event_name).map(self._call_one)
    124     def _call_one(self, event_name):

~/Projects/fastcore/fastcore/foundation.py in map(self, f, *args, **kwargs)
    361              else f.__getitem__)
--> 362         return self._new(map(g, self))
    363 

~/Projects/fastcore/fastcore/foundation.py in _new(self, items, *args, **kwargs)
    314     def _xtra(self): return None
--> 315     def _new(self, items, *args, **kwargs): return type(self)(items, *args, use_list=None, **kwargs)
    316     def __getitem__(self, idx): return self._get(idx) if is_indexer(idx) else L(self._get(idx), use_list=None)

~/Projects/fastcore/fastcore/foundation.py in __call__(cls, x, *args, **kwargs)
     40 
---> 41         res = super().__call__(*((x,) + args), **kwargs)
     42         res._newchk = 0

[...]

PR #378, opened 2020-05-26
ShowGraphCallback needs remove_on_fetch=True

indigoviolet:
I think this is a similar issue to #156 -- I get the same error about the Learner missing a Recorder if I'm using `ShowGraphCallback` with `SaveModelCallback` (and `WandbCallback`).


sgugger:
SGTM, thanks!

PR #379, opened 2020-05-26
fixes a typo in callback schedule test

ldanilov:
Looks like just a typo. 

But another cell in this notebook is also failing: 
`#Integration test: training a few epochs should make the model better`


sgugger:
The other cell you mention fails randomly sometimes. We should add a seed there to make it always pass.

PR #381, opened 2020-05-26
Vision transforms improvements

antoinebon:
Extend `Zoom` transform with zoom out capabilities with `min_zoom`
RandomRsizedCrop needs at least order==1 to run after PointScaler
Resize.final_size is not used, so I removed it for code clarity


sgugger:
Why not, though I'm very doubtful min_zoom < 1 could be useful (you only add padding). Thanks!

antoinebon:
Thanks! Could be useful to train conv kernels to operate on lower scales

Issue #373, opened 2020-05-22
Side effect with crop_pad on TensorPoint


antoinebon:
As illustrated in the following gist https://gist.github.com/antoinebon/368ec4a5a996eefd179edfaced7db18d, resizing a TensorPoint with crop_pad replicates the original TensorPoint's `__dict__` (in `Tensor.set_meta`). Therefore, changing the `_meta` of the resized TensorPoint also also changes the `_meta` of the orignal TensorPoint

This side effect is not very transparent and probably not desired. What do you think?

sgugger:
Eh, no, it doesn't look like something we want. Will try to come up with a fix.
Thanks for the reproducer!

muellerzr:
Since this was merged with a fix, can this issue be closed @antoinebon ?

antoinebon:
@muellerzr I don't think this issue is fixed, at least it was not fixed in #382.

jph00:
Fixed now!

PR #382, opened 2020-05-26
crop_pad improvements for TensorPoint

antoinebon:
`crop_pad` failed for TensorPoint in the situation where `orig_sz` or `tl` are `None` because `TensorPoint` has no `size` attribute (which happens in `RandomResizeCrop.encodes` for validation split)
I also extened `_get_sz` to search additionally for `sz` meta (which I think is set during casting to TensorPoint). This happens when calling `crop_pad` twice in a row, which is the case in `RandomResizeCrop.encodes` for validation split (somewhat related to issue https://github.com/fastai/fastai2/issues/373)


sgugger:
Looks good to me, thanks!

PR #383, opened 2020-05-27
add MatthewsCorrCoef, PearsonCorrCoef, SpearmanCorrCoef

richardyy1188:
First, I only modify `metrics.ipynb` and then run all, I have no idea why so many files are modified. I think you will be clearer on what is happened if be able to see what is changed, so I still opened this PR.  :joy:

This PR is based on this [thread](https://forums.fast.ai/t/pr-discuss-add-pearson-and-spearman-corref-metric/71934/4).
I add with minimum modification on original code.

But there is one thing to notice. Although there is `axis` argument in `scipy.stats.spearmanr` but we won't be able to specify other value but just use default value, because it conflict with `axis` in `skm_to_fastai`, we need to change `axis` in `skm_to_fastai` to other name to fix this.(Notice it still work when we want to use default value of axis of spearman’s)

BTW, since we also use scipy, maybe we should rename `skm_to_fastai` to other name ?


fastai2/metrics.py
```diff
@@ -169,6 +172,23 @@ def RocAucBinary(axis=-1, average='macro', sample_weight=None, max_fpr=None, mul
     return skm_to_fastai(skm.roc_auc_score, axis=axis, activation=ActivationType.BinarySoftmax,
                          average=average, sample_weight=sample_weight, max_fpr=max_fpr, multi_class=multi_class)
 
+# Cell
+def MatthewsCorrCoef(sample_weight=None, **kwargs):
+    "Matthews correlation coefficient for single-label classification problems"
+    return scim_to_fastai(skm.matthews_corrcoef, sample_weight=sample_weight, **kwargs)
```
sgugger, line 178: The function `scim_to_fastai` is not defined.

sgugger:
Thanks for the PR. Like I said on the thread, I'd rather stop adding arguments to `skm_to_fastai` or change their names since that function is just three lines of codes and is starting to get too complex. We can just write another function, similar, with three lines of codes too and change args there to be easier to use with scipy ;)

You need to be using nbdev master branch right now since there have been big changes and fixes, until we make the next release. That's why your diff is so large.

PR #389, opened 2020-06-01
Remove unused try-except Block for easier Debugging

flpeters:
When creating a DataLoader, this try-except catches any error messages and simply prints that something went wrong. The DataLoader is most likely unusable at this point, so in my opinion there is no need for a "soft" failure here.
By removing it, one can see what actually went wrong and fix it.


jph00:
We actually use this feature quite a bit.

PR #391, opened 2020-06-07
PearsonCorrCoef and SpearsonCorrCoef are for regression problems

richarddwang:
PearsonCorrCoef and SpearsonCorrCoef are for regression problems and thus don't use argmax.


sgugger:
HI there, thanks for your PR!
Please note that the default for `dim_argmax` in `AccumMetric` is already `None`, so you don't need to pass that value once you remove the default in `scs_to_fastai`.

Please keep all docstrings to one line and add any more comments in a markdown cell in the notebook.

richardyy1188:
Hi, is there anything new ? Cuz I think we should merge this to fix that bug.

PR #392, opened 2020-06-07
Fixes test in callback schedule

ldanilov:
Previous test failed, because running fit resets the loss to None.


sgugger:
Thanks for the fix!

PR #394, opened 2020-06-07
Fixes exception message to use compatible version of sentencepiece lib

ldanilov:
`nbdev_test_nbs --fname nbs/30_text.core.ipynb`

Returns:`OSError: Not found: unknown field name "minloglevel" in TrainerSpec.` 

Because it looks like `minloglevel` flag was dropped from version 0.1.90 and 0.1.91 of sentencepiece library. 

But it will be back in 0.1.92:
https://github.com/google/sentencepiece/commit/5bf131908d33b3fdf626f4f3972fe83fc4da1605#diff-20a88935144fc7c1f34aa848fec8cc60R145-R155


sgugger:
Seriously?
Thanks for fixing the install instructions!

PR #397, opened 2020-06-13
Update comments: specify process relationship

lsb:
I believe this clarifies existing comments, about not saving in a forked child process.

I believe this does not fall under the fastai1 `contributing.md` provision against Parkinson's law of triviality, as it both helps clarify language and reduces violent metaphors in code comments.

(For a larger project of moving the latest commits onto a branch called latest, which would require some community heads-up, something like https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx )


sgugger:
Yes, I agree those terms are better.

PR #400, opened 2020-06-15
F.interpolate compatability patch for pytorch 1.5.0

streicherlouw:
F.interpolate, used by the RatioResize transform in fastai2’s batch image augmenter, changed it’s default behaviour between pytorch versions v1.4.0 and v1.5.0, causing installations using pytorch v1.5.0 (such as google colab) to throw a warning that using F.interpolate in this way will fail in future. Adding the argument “recompute_scale_factor=True” when calling F.interpolate prevents pytorch from throwing this warning, and ensures fastai compatibility with future versions of pytorch.

### DIFF
--- a/fastai2/vision/augment.py
+++ b/fastai2/vision/augment.py
@@ -285,7 +285,7 @@ def _grid_sample(x, coords, mode='bilinear', padding_mode='reflection', align_co
         d = min(x.shape[-2]/coords.shape[-2], x.shape[-1]/coords.shape[-1])/2
         # If we're resizing up by >200%, and we're zooming less than that, interpolate first
         if d>1 and d>z:
-            x = F.interpolate(x, scale_factor=1/d, mode='area')
+            x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
     return F.grid_sample(x, coords, mode=mode, padding_mode=padding_mode, align_corners=align_corners)
 
 # Cell
--- a/fastai2/vision/augment.py
+++ b/fastai2/vision/augment.py
@@ -285,7 +285,11 @@ def _grid_sample(x, coords, mode='bilinear', padding_mode='reflection', align_co
         d = min(x.shape[-2]/coords.shape[-2], x.shape[-1]/coords.shape[-1])/2
         # If we're resizing up by >200%, and we're zooming less than that, interpolate first
         if d>1 and d>z:
-            x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
+            # Pytorch > v1.4.x needs an extra argument when calling nn.functional.interpolate to preserve previous behaviour
+            if (int(torch.__version__[0:3].translate({ord('.'): None})) > 14):
+                x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
+                else:
+                x = F.interpolate(x, scale_factor=1/d, mode='area')
     return F.grid_sample(x, coords, mode=mode, padding_mode=padding_mode, align_corners=align_corners)
 
 # Cell
--- a/fastai2/vision/augment.py
+++ b/fastai2/vision/augment.py
@@ -286,7 +286,7 @@ def _grid_sample(x, coords, mode='bilinear', padding_mode='reflection', align_co
         # If we're resizing up by >200%, and we're zooming less than that, interpolate first
         if d>1 and d>z:
             # Pytorch > v1.4.x needs an extra argument when calling nn.functional.interpolate to preserve previous behaviour
-            if (int(torch.__version__[0:3].translate({ord('.'): None})) > 14):
+            if (int(torch.__version__[0:4].replace(".", "")) > 14):
                 x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
             else:
                 x = F.interpolate(x, scale_factor=1/d, mode='area')
--- a/fastai2/vision/augment.py
+++ b/fastai2/vision/augment.py
@@ -288,7 +288,7 @@ def _grid_sample(x, coords, mode='bilinear', padding_mode='reflection', align_co
             # Pytorch > v1.4.x needs an extra argument when calling nn.functional.interpolate to preserve previous behaviour
             if (int(torch.__version__[0:3].translate({ord('.'): None})) > 14):
                 x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
-                else:
+            else:
                 x = F.interpolate(x, scale_factor=1/d, mode='area')
     return F.grid_sample(x, coords, mode=mode, padding_mode=padding_mode, align_corners=align_corners)
 


jph00:
Does it still work on 1.4.0 with this parameter added?

streicherlouw:
Hi Jeremy.

The PR400 change, as proposed, does *not* work on pytorch 1.4.0. I have been in communication with @muellerzr about how to resolve this for the past few days, and here is a summary of where we have gotten to so far:

**The problem**

Fastai2, as is, throws a warning (but continues to work) on pytorch 1.5.0 because the scale factor is passed to F.interpolate a float in RatioResize on line 289 of augment.py:
```
class RatioResize(Transform):
    'Resizes the biggest dimension of an image to `max_sz` maintaining the aspect ratio'
....
def _grid_sample(x, coords, mode='bilinear', padding_mode='reflection', align_corners=None):
    "Resample pixels in `coords` from `x` by `mode`, with `padding_mode` in ('reflection','border','zeros')."
    #coords = coords.permute(0, 3, 1, 2).contiguous().permute(0, 2, 3, 1) # optimize layout for grid_sample
    if mode=='bilinear': # hack to get smoother downwards resampling
        mn,mx = coords.min(),coords.max()
        # max amount we're affine zooming by (>1 means zooming in)
        z = 1/(mx-mn).item()*2
        # amount we're resizing by, with 100% extra margin
        d = min(x.shape[-2]/coords.shape[-2], x.shape[-1]/coords.shape[-1])/2
        # If we're resizing up by >200%, and we're zooming less than that, interpolate first
        if d>1 and d>z:
289>>            x = F.interpolate(x, scale_factor=1/d, mode='area')
    return F.grid_sample(x, coords, mode=mode, padding_mode=padding_mode, align_corners=align_corners)
```

The problem manifests as a message stating ``UserWarning: The default behavior for interpolate/upsample with float scale_factor will change in 1.6.0 to align with other frameworks/libraries, and use scale_factor directly, instead of relying on the computed output size. If you wish to keep the old behavior, please set recompute_scale_factor=True. See the documentation of nn.Upsample for details``

The warning seems to be simply informing of an upcoming change, as the pytorch code simply checks if scale_factor is a float or not ([torch/nn/functional.py](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py) line 2989):
```
 if recompute_scale_factor is None:
        # only warn when the scales have floating values since
        # the result for ints is the same with/without recompute_scale_factor

        is_float_scale_factor = False
        for scale in scale_factors:
            is_float_scale_factor = math.floor(scale) != scale
            if is_float_scale_factor:
                break
2989>>   if is_float_scale_factor:
            warnings.warn("The default behavior for interpolate/upsample with float scale_factor will change "
                          "in 1.6.0 to align with other frameworks/libraries, and use scale_factor directly, "
                          "instead of relying on the computed output size. "
                          "If you wish to keep the old behavior, please set recompute_scale_factor=True. "
                          "See the documentation of nn.Upsample for details. ")
```
The major risk of this warning is not that fastai2 fails on pytorch 1.5.0 (it works fine, it just *looks* ominous), but the confusion that the warning causes to people unfamiliar with fastai. [This fast.ai forum post](https://forums.fast.ai/t/fastai-throwing-a-runtime-error-when-using-custom-train-test-sets/70262) is a good example of a conversation that stated out discussing a valid problem with a dataset but ended up discussing the dataloader error message because of the pytorch warning.

It also makes people [think that they have to downgrade](https://forums.fast.ai/t/fastai-throwing-a-runtime-error-when-using-custom-train-test-sets/70262/3?u=streicher) the pytorch in colab before they can use fastai2 which is untrue, wasteful, and incoherent with how cutting edge fastai2 truly is.

**Fixes** 
As far as I can work out, there are two ways to fix this:

A: Add "recompute_scale_factor=True" to the F.interpolate call (PR400) or
B: Use "size" in stead of "scale_factor" in the F.interpolate call (probably a better option)

**Solution A: Add "recompute_scale_factor=True" to the F.interpolate call**

PR400 takes it's cue from the [pytorch documentation for torch.nn.functional.interpolate (aka nn.Upsample)](https://pytorch.org/docs/stable/nn.functional.html?highlight=interpolate#interpolate) that elaborates as follows:

`recompute_scale_factor (bool ,optional) – recompute the scale_factor for use in the interpolation calculation. When scale_factor is passed as a parameter, it is used to compute the output_size. If recompute_scale_factor is True or not specified, a new scale_factor will be computed based on the output and input sizes for use in the interpolation computation (i.e. the computation will be identical to if the computed output_size were passed-in explicitly). Otherwise, the passed-in scale_factor will be used in the interpolation computation. Note that when scale_factor is floating-point, the recomputed scale_factor may differ from the one passed in due to rounding and precision issues.`

This fix worked for me as I use pytorch 1.5.0 on all my devices with fastai2, but on subsequent testing with pytorch 1.4.0 and 1.3.0 it unfortunately failed as follows:
```
>>> torch.__version__
'1.4.0'
> /home/linuxbrew/.linuxbrew/Cellar/python/3.7.7_1/lib/python3.7/site-packages/fastai2/vision/augment.py(289)_grid_sample()
-> x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
(Pdb) s
TypeError: interpolate() got an unexpected keyword argument 'recompute_scale_factor'
```
Interestingly, on <1.5.0 it failed almost silently as the error message above was not propagated to the python command line or jupyter (hence the debug in pdb). The only sign that there was a problem was a fairly cryptic error message from fastai that it "Could not do one pass in your dataloader, there is something wrong in it" whenever aug_transforms was used.

One could consider including a check for pytorch version before doing the F.interpolate call, but a cleaner option may be to simply pass size instead of scale factor to F.interpolate.

**Solution B: Use "size" in stead of "scale_factor" in the F.interpolate call**

As the "size" argument in torch.nn.functional.interpolate is stying constant across pytorch versions,  passing F.interpolate size instead of scale factor could provide an elegant way to eliminate the warning while remaining pytorch-version agnostic. Looking at the code around lines 280 to 288 of [augment.py](https://github.com/fastai/fastai2/blob/master/fastai2/vision/augment.py) suggests to me that this should be doable, but I do not understand the workings of the RatioResize routine well enough to work out how to do it, or more importantly, how to test if any code I come up with does not break something else. 

May I ask if you could investigate if passing "size" to F.interpolate is a better option than my proposed solution of adding recompute_scale_factor and a version check?

**PS.**
(PS. Thank you so much for the early information on masks during the fastai2 course. Your data and videos directly inspired my 76 yo mother, who lives in South Africa, to take the store of fabric she had built up over a lifetime and produce stacks of masks that were distributed to the informal settlements around where she lives. These continue to provide protection for people who are only now going through the worst of their pandemic. You are unequivocally one of my heroes.)

sgugger:
I think there is no other choice than to test the pytorch version and have two different calls to `F.interpolate` depending on the result...

streicherlouw:
Thank you for the quick response response Sylvain. I will update PR400 with a test statement and resubmit.

streicherlouw:
I added a check that uses ``torch.__version_`` to divine whether or not to insert the additional argument in the F.interpolate call.

As ``torch.__version_`` is sometimes adulterated with additional characters like ``1.5.0+cu101`` on platforms like google colab, the check uses only the first four characters, which are stripped of punctuation and converted to int, so that ``(string)1.5.0+cu101`` becomes ``(string)1.5.`` which becomes ``(string)15`` which becomes ``(int)15``. Using the first *four* characters ensures that the resulting int from versions ``2.0.0`` (20)  and ``1.10.0`` (100) are both greater than the resulting int from version ``1.4.0`` (14), which is used for the check.

I also timed several ways to do the conversion, with 100 000 runs of string.translate and string.replace resulting in execution times of 0.71s and 0.65s respectively.
 
``(int(torch.__version__[0:4].translate({ord('.'): None})) > 14)``  : 0.7124877270016441s
``(int(torch.__version__[0:4].replace(".", "")) > 14)``  : 0.6456794480000099s

The final new codeblock thus looks as follows:

```
if (int(torch.__version__[0:4].replace(".", "")) > 14):
    x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
else:
    x = F.interpolate(x, scale_factor=1/d, mode='area')
```
The latest commit has been successfully tested on pytorch v1.4.0, v1.5.0 and google colab running 1.5.0+cu101
```
(Pdb) torch.__version__
'1.4.0'
(Pdb) s
> /home/streicher/fastai2/fastai2/vision/augment.py(289)_grid_sample()
-> if (int(torch.__version__[0:4].replace(".", "")) > 14):
(Pdb) s
> /home/streicher/fastai2/fastai2/vision/augment.py(292)_grid_sample()
-> x = F.interpolate(x, scale_factor=1/d, mode='area')
```
```
(Pdb) torch.__version__
'1.5.0'
(Pdb) s
> /home/streicher/fastai2/fastai2/vision/augment.py(289)_grid_sample()
-> if (int(torch.__version__[0:4].replace(".", "")) > 14):
(Pdb) s
> /home/streicher/fastai2/fastai2/vision/augment.py(290)_grid_sample()
-> x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
```
This PR should now be ready for merge.

jph00:
Many thanks for this! :)

PR #402, opened 2020-06-17
Small fix to docstring for the SentencePiece tokenizer

davanstrien:
The SentencePiece tokenizer currently refers to Spacy' in the docstring instead of SentencePiece. This isn't a huge deal but might cause some confusion.


sgugger:
Thanks!

PR #403, opened 2020-06-17
Save Model Callback: print new best score and epoch id on improvement

riccardoangius:
First off, thank you all for your work! As a CS/AI student, it is immensely convenient to just be able to concentrate on the higher level stuff.

I am opening this PR to introduce a print for the new best score and epoch id, after each epoch, when a Save Model Callback is added to the model and set to save on every improvement.

It essentially ports the alert as present in v1.


sgugger:
I don't see why not, thanks for your PR!

PR #405, opened 2020-06-21
Add backwards support for text models and MixedDL

muellerzr:
Should be all that's needed, let me know if you see anything wrong with the implementation 

When testing on IMDB_Sample improvement was seen within the ensemble


muellerzr:
I meant to put this in a separate PR however, this also introduces MixedDL, a process for multi-modal models to easily merge `DataLoaders` together. It takes care of the various repeats that could potentially happen, while also being efficient and easy to use

jph00:
Thanks! I'm merging this, but can you please do another PR when you have a chance to remove MixedDL? I'd like to better understand what this is for, and check whether this is the best way to use it, before including it in fastai.

muellerzr:
Sure thing! The MixedDL is a general method for multi-modal data techniques

PR #411, opened 2020-06-24
CorpusBLEU smoothing and tests

morganmcg1:
Adding: 1) Smoothing to account for instances where no n-gram matches are found. Followed method used in SacreBLEU. 2) tests for CorpusBLEU


jph00:
Thanks!

PR #418, opened 2020-07-09
feat(wandb): add support for tabular data

borisdayma:
Add support for tabular data to `WandbCallback`.

See [example run](https://app.wandb.ai/borisd13/demo_config/runs/27km4shg?workspace=user-borisd13).

Thanks a lot to @muellerzr for the help on this!


jph00:
Many thanks. Would it make sense to move `gather_args` into wandb nb, since it's only used there?

borisdayma:
The idea was that it could also be used for general debug, mainly to check which args were passed to each sub-function (and sometimes also detect signature issue).
However I can move it to wandb nb if you prefer.

PR #420, opened 2020-07-09
Modifies SortedDL.new for val_res

morganmcg1:
Modifies SortedDL.new to accept val_res, as list of validation set text lengths, to further speed up the init. Also adds additional documentation for SortedDL to emphasise the benefit of passing res to the dataloader


morganmcg1:
Further elaborated on here: https://forums.fast.ai/t/nlp-speed-up-if-using-sorteddl/74636

jph00:
Thank you!

Issue #417, opened 2020-07-04
'_FakeLoader' object has no attribute 'generator'


jekriske-lilly:
Granted pytorch 1.6.0 hasn't been released yet, I'm seeing this with the latest pytorch build and the latest fastai2. 

If this is simply a pytorch error then possibly ignore. Curious why nbdev_test_nbs didn't pick this up though...

```
coll = range(15)
dl = DataLoader(coll, batch_size=5, shuffle=True)
list(dl)
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-93-32442226efef> in <module>
      1 coll = range(15)
      2 dl = DataLoader(coll, batch_size=5, shuffle=True)
----> 3 list(dl)

~/src/fastai2/fastai2/data/load.py in __iter__(self)
     96         self.randomize()
     97         self.before_iter()
---> 98         for b in _loaders[self.fake_l.num_workers==0](self.fake_l):
     99             if self.device is not None: b = to_device(b, self.device)
    100             yield self.after_batch(b)

~/fastai_venv/lib/python3.7/site-packages/torch/utils/data/dataloader.py in __init__(self, loader)
    383 class _SingleProcessDataLoaderIter(_BaseDataLoaderIter):
    384     def __init__(self, loader):
--> 385         super(_SingleProcessDataLoaderIter, self).__init__(loader)
    386         assert self._timeout == 0
    387         assert self._num_workers == 0

~/fastai_venv/lib/python3.7/site-packages/torch/utils/data/dataloader.py in __init__(self, loader)
    339         self._collate_fn = loader.collate_fn
    340         self._sampler_iter = iter(self._index_sampler)
--> 341         self._base_seed = torch.empty((), dtype=torch.int64).random_(generator=loader.generator).item()
    342         self._num_yielded = 0
    343 

AttributeError: '_FakeLoader' object has no attribute 'generator'
```

davidrichards:
I think you're right, I think it's a PyTorch version issue. I couldn't reproduce the issue:

```
from fastai2.basics import *
coll = range(15)
dl = DataLoader(coll, batch_size=5, shuffle=True)
list(dl)
```

Gives me:

```
[tensor([11,  3,  7,  6, 10]),
 tensor([ 2, 14,  5,  8, 13]),
 tensor([ 4,  0, 12,  1,  9])]
```

[...]

PR #426, opened 2020-07-27
DataLoader fix for PyTorch nightly

tmabraham:
This commit adds necessary changes in order to solve #417. This would allow the fastai2 DataLoaders to work with PyTorch nightly.

I will note that this hard-codes the `generator` and `prefetch_factor` to the default PyTorch DataLoader arguments, and I am not sure if this is necessarily wanted. If Jeremy has a different approach in mind, we can discuss and I can close this pull request.

Also, this is my first time contributing to fastai2, so please let me know if there's some mistakes I may be making.


jph00:
Great - thanks!

PR #427, opened 2020-07-27
Fix bug on missing backpropagation in WeightDropout

riven314:
In ```WeightDropout```, it applies dropout mask on target weight specified in  ```layer_names```. Current implementation accidentally disables target weights to do back-propagation. Here is a simple code to show the bug:  
```
import torch
import torch.nn as nn
from fastai2.text.models.awdlstm import WeightDropout

lstm = nn.LSTM(3, 5, batch_first = True)
# target weight: weight_hh_l0
# non-target weight: weight_ih_l0
lstm_dp = WeightDropout(lstm, weight_p = 0.8, layer_names='weight_hh_l0')

test_input = torch.randn(8, 20, 3)  # (batch size, seq length, input dim)
test_h = torch.randn(1, 8, 5)
test_c = test_h.data
output, (h, c) = lstm_dp(test_input, (test_h, test_c))

loss = output.sum()
loss.backward()

#show that non-target weight have gradients computed but target weight hasn't
assert lstm_dp.module.weight_ih_l0.grad is not None 
assert lstm_dp.weight_hh_l0_raw.grad  is None
```

This PR makes two changes:
1. to fix the above-mentioned bug
2. include additional tests on ```nbs/32_text.models.awdlstm.ipynb``` to verify that gradients are computed for target weights


jph00:
This is great! Thanks so much :)

PR #438, opened 2020-07-30
Adjust name properties to transforms, move to classes, and print parameters better

muellerzr:
This is the first part of setting up the `.name`'s for transforms to make them more readable. An example output is below:
```python
FlipItem: {'p': 1.0}: (TensorBBox,object) -> encodes
(TensorPoint,object) -> encodes
(TensorMask,object) -> encodes
(TensorImage,object) -> encodes
(Image,object) -> encodes 
```
From here I'll be moving to tabular, then data (Categorize, Normalize), followed by text.


muellerzr:
It may be better to have Transform inherit this though, let me play around with that

muellerzr:
I've added tabular and data, I don't believe that text may be worth it, as those two transforms aren't usually customized much. (Numericalize and Tokenizer)

Issue #442, opened 2020-07-31
Moving docs raising issues with nbdev package


danieldiamond:
## Description
https://github.com/fastai/fastai2/blob/a784012254cd52abede0c4c1a111940d1cce7478/docs#L1

Changes from [latest commits ](https://github.com/fastai/fastai2/commit/6801f91e2eab992c02d4405ec478cdc3181e12b1) are causing issues

`nbdev_build_lib` is calling [write_tmpls](https://github.com/fastai/nbdev/blob/master/nbdev/export2html.py#L500) but these files have been reorganized / the latest commits seem to remove the docs dir entirely

## Steps to Reproduce
1. clone repo and set up repo per `contributions` in Readme
2. `nbdev_build_lib`or `make test`

muellerzr:
Yes, you need to clone the `fastai_docs` repo inside of your `fastai2` repo. It sym-links

Issue #439, opened 2020-07-30
sklearn's `matthews_corrcoef` doesn't really support multi-label, hence `MatthewsCorrCoefMulti` won't work


tianjianjiang:
```python
/usr/local/lib/python3.6/dist-packages/sklearn/metrics/_classification.py in matthews_corrcoef(y_true, y_pred, sample_weight)
    883     check_consistent_length(y_true, y_pred, sample_weight)
    884     if y_type not in {"binary", "multiclass"}:
--> 885         raise ValueError("%s is not supported" % y_type)
```

`ValueError: multilabel-indicator is not supported`

jph00:
This isn't enough information for me to understand. Please provide a minimal reproducer, say what you expect to happen, and show what actually happened.

tianjianjiang:
> This isn't enough information for me to understand. Please provide a minimal reproducer, say what you expect to happen, and show what actually happened.

My apologies for the trouble. Judging by the code of sklearn, there's no way it can do multi-label `matthews_corrcoef()`, therefore I didn't provide any reproducer. To be more specific:

1. `MatthewsCorrCoefMulti` uses `skm_to_fastai`
2. `skm_to_fastai` uses `skm.matthews_corrcoef`, i.e., `sklearn.metrics.matthews_corrcoef`
3. `sklearn.metrics.matthews_corrcoef` uses `y_type, y_true, y_pred = _check_targets(y_true, y_pred)`
4. `_check_targets` correctly detects `y_type` as `multilabel-indicator`
5. `sklearn.metrics.matthews_corrcoef` rejects `multilabel-indicator` @ https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/_classification.py#L834

For your reference, I bumped into this when I was reproducing https://ohmeow.github.io/blurr/modeling-core/#Example:-Multi-label-classification with additional metrics including `MatthewsCorrCoefMulti`.

jph00:
Thanks. So would your suggestion be for us to remove that function?

danieldiamond:
Enjoyed following this turn of events:

It appears it was previously removed [here](https://github.com/fastai/fastai2/commit/4bde87b0a2144f6ad4e816ad4c2ab48628b911a7) -- there was a previous issue raised [here](https://github.com/fastai/fastai2/issues/155)

But then it was added again [here](https://github.com/fastai/fastai2/pull/384)

[...]

PR #443, opened 2020-07-31
Removing matthewscorrcoef & Restructure Docs

danieldiamond:
## Description
- removing matthewscorrcoef per https://github.com/fastai/fastai2/issues/439
- attempting to resolve docs issue https://github.com/fastai/fastai2/issues/442
    - removed symlink `docs` and added empty docs dir.
    - added the appropriate `_data/topnav.yaml` file
    - `_config.yaml` is repopulated although I imagine you do not want this included in the PR
    - Trying to reason the changes made in `move docs` commits, will need some input/insight into how to resolve the above issue (442).

## Comments
- `make test` fails due to kernel naming issues (tests running notebooks require python3 environment)
- It appears `jupyter_client` was also missing as a requirement for testing notebooks
- `02a_data.mixed.ipynb` is being removed when running `nbdev_update_lib`..


jph00:
I'll close this and fix locally, since we don't want those docs changes.

PR #466, opened 2020-08-05
Allow repo admins to trigger tests manually from the Actions Tab in Github for this repo

hamelsmu:
@jph00 this gives you the ability to manually trigger workflows from the Actions tab.  Only admins (you) will be able to do this.

It looks like this 

image

Helpful for the future in case there is an error with the environment or Docker so you can manually trigger the workflow if you wish.


jph00:
Very interesting. I should move some stuff out of the Makefile and into nbdev I suspect then!

PR #471, opened 2020-08-06
Create welcome.yml

hamelsmu:


hamelsmu:
cc: @jph00 merging this welcome template will monitor and see if it works

PR #480, opened 2020-08-07
mark LSUN dataset and all associated steps as slow

hamelsmu:
@jph00 as discussed marked these steps as slow.  I am not sure why GitHub Actions is not ignoring the slow cells yet, I think it may be checking out the base branch and not the head of this branch.

I am going to have to investigate that issue separately in a toy repo and do some experiments.


jph00:
Easier would be to just put a single cell containing "# all_slow" at the top. That will mark the whole NB as slow.

hamelsmu:
@jph00 This is done now

PR #489, opened 2020-08-07
Revert README.md back to original state

hamelsmu:


jph00:
Don't change README - instead change index.ipynb

PR #511, opened 2020-08-08
remove **kwargs in load_image

KushajveerSingh:
I checked the PIL.Image.open documentation. There is only one argument `mode='r'`. In the documentation it says `The mode. If given, this argument must be “r”.`, which is the default value.


hamelsmu:
Hi @KushajveerSingh I patched a potential security vulnerability to CI so I am rebasing your branch w/latest changes

PR #518, opened 2020-08-10
Pearson and Spearman Correlation are mainly used for regression problem.

richarddwang:
Currently the two metric is `dim_argmax=-1`, but they are mainly used for regression problem (e.g. GLUE's stsb task), which pred is not logits but scores with the same shape with targets.

I move the two to Regression section, and make default `dim_argmax=None` for them, but users can still specify `dim_argmax=-1` if they want to use them in classification problem.


jph00:
Seems you removed `scs_to_fastai`? Could you please add tests for these when you move them?

richarddwang:
Tests for Pearson and Spearman correlation coefficient has been added.

jph00:
Thanks! I think `scs_to_fastai` is still removed though?

richarddwang:
Seems that I misunderstand something.

I add `scs_to_fastai` with `PearsonCorrCoef` and `SpearmanCorrCoef` metrics in a pr. And now I found `scs_to_fastai` is actually not useful and redundant so I remove it.

Is there any things I need to do now ?

jph00:
Yes please put it back. fastai has a public API. Only things with a `_` suffix are private API. We don't remove things from the public API if it's not used internally - users are still using it externally.

jph00:
...well, I guess it's simple enough that maybe it's not a problem. I'll merge the PR, and if folks complain that they were using it, we can always put it back :)

richarddwang:
Oh, I see. 
And thank you !

PR #520, opened 2020-08-10
Let prediction of MSELoss be able to be bounded

richarddwang:
For example, stsb task in GLUE benchmark predicts the sentiment of a sentence within **[0., 5.]**. If we don't bound the prediction, it can be out of interval like -0.098.  

So I made `MSELossFlat` be able to set 2 arguments `ceil` and `floor`, so `MSELossFlat(floor=0.0, ceil=5.0)` will make prediction < 0. to 0.,and > 5. to 5. to solve the problem.
(the two arguments are default to None)

I have succeed to replicate extactly STSB test score (84.6) of ELECTRA-small++ with this twitch, so it should be good.


hamelsmu:
@richarddwang I patched a security vulnerability with CI so I will update your branch to include the latest changes.

jph00:
We're not adding new functionality until after release - feel free to send in this PR then, although when you do, please include documentation and tests in your PR.

PR #523, opened 2020-08-10
feat(wandb): use artifacts to track datasets

borisdayma:
Datasets can now be tracked by using `WandbCallback(dasaset_path='my_data')`

It uses [W&B artifacts](https://docs.wandb.com/artifacts).

See [example artifacts](https://app.wandb.ai/borisd13/fastai2_artifacts/artifacts) on camvid dataset.

The dataset is versioned and has aliases such as `latest` (custom ones can be used) and will identify any changes of dataset between runs.
Different datasets can be used by using different names.

Here is a [demo notebook](https://gist.github.com/borisdayma/6bf35b9fd8126f97742ec9b1c0b7110e).

**TODO in subsequent PR:**
* trained models should also be logged as artifacts for better tracking and to allow more flexible pipelines (instead of files uploaded from a run)


borisdayma:
Note: I'm using master branch of nbdev. Not sure if I should use the last released version instead for CI

jph00:
Yes last released version is best - thanks!

PR #527, opened 2020-08-12
feat(wandb): improve dataset logging

borisdayma:
Changes:
* arg `dataset_path` becomes `log_dataset`
* if set to `True`, it uses `dls.path` but it can also explicitly be set to a path
* "models" subfolder is always ignored (I don't see how it could be in any valid dataset)
* the function `log_dataset` is now public so people can manually do something more complex (such as logging train and valid separately)
* added documentation related to tracking datasets


borisdayma:
Changes:
* `metadata={'path': …}` is now replaced with `metadata={'path relative to learner`:…}`
* function `log_dataset` now accepts argument `learner_path`
* `learner_path` is automatically set during the callback ; if we call `log_dataset` manually and don't set `learner_path`, metadata will not be logged

Note: `os.path.relpath` can return results such as `..\data` while `Path.relative_to` fails

PR #536, opened 2020-08-16
feat(wandb): log models as artifacts

borisdayma:
**Changes:**

1. removed `SaveModelCallback.add_save` as it is not used anymore
1. `learn.save` now returns a path
1. use `SaveModelCallback.last_path` to keep track of last model saved for access to loggers. This was the most simple approach I thought of.
1. add `log_model` method to upload model as artifact
1. cleaner approach to `log_model` and `log_dataset` where we can directly pass the metadata we want (automatically generated when using `WandbCallback`)
1. we still know which model to log through the presence of `SaveModelCallback` (custom metric…) but we can now also desactivate model logging with `log_model=False` (eg when doing parameter search). Non-breaking change as it defaults to `log_model=True`

See [example](https://app.wandb.ai/borisd13/fastai2_artifacts/artifacts/model/run-2bafen4k-model/41ab3e7cfe740ce5fe4e4993c569865c/graph) and [demo notebook](https://gist.github.com/borisdayma/6bf35b9fd8126f97742ec9b1c0b7110e).

If you like the approach, feel free to merge and I'll update doc in follow-up PR.


jph00:
Can you please re-post this at the new fastai/fastai repo? Sorry for the inconvenience!

PR #537, opened 2020-08-17
Export to library the import of Image

albertvillanova:
Library file `torch_core.py` uses `PIL.Image` in the definition of the function `to_image`. However it was not previously imported.

Now, `torch_core.py` includes the import:
```
from PIL import Image
```


jph00:
I'll try to find a fix for that too. I'd rather not intro the dep.

PR #538, opened 2020-08-17
Add ipython as requirement

albertvillanova:
The package `ipython` is not installed when using `pip`.

However, it is used by function `torch_core.display_df`.

Maybe, additionally both seutup.py requirements and conda environment.yml should be aligned.


jph00:
I'll try to find a way to remove that dep.

